PR #17, opened 2018-01-10
Set devise secret on container build

yssavchuk:
- [x] Different device secret key for each rails environment
- [x] Use default device secret key for container build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,11 @@ FROM ruby:2.5.0
 #
 ARG RAILS_ENV=production
 ENV RAILS_ENV ${RAILS_ENV}
-ENV DEVICE_SECRET_KEY='changeme'
+
+# Devise requires secret key to be set during image build or it raises an error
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -9,6 +9,7 @@ FROM ruby:2.5.0
 #
 ARG RAILS_ENV=production
 ENV RAILS_ENV ${RAILS_ENV}
+ENV DEVICE_SECRET_KEY='changeme'
```
yivo: @yssavchuk Could you please add comment why this is needed: `Devise requires secret key to be set or it raises an error preventing from running any scripts. Users should override this variable by passing environment variable to Docker`.

yivo:
@mod It is used for devise and should be namespaced to understand what it is used for.

yivo:
SECRET_KEY is just confusing.

yivo:
@mod We have updated from DEVICE to DEVISE!

PR #18, opened 2018-01-10
Front-end: Bootstrap Gem, Devise forms, Skinning

Mons-web:


yivo:
@Mons-web Please, rebase from master. I will help you to get everything work in this PR.

app/assets/javascripts/admin.js
```diff
@@ -1,4 +1,4 @@
 //= require rails-ujs
 //= require jquery3
 //= require popper
-//= require bootstrap-sprockets
```
yivo: @Mons-web We should always prefer to use Sprockets-mastered script since Rails uses Sprockets :) Also this is the recommendation of the gem itself https://github.com/twbs/bootstrap-rubygem

app/assets/javascripts/web.js
```diff
@@ -1 +1,4 @@
 //= require rails-ujs
+//= require jquery3
+//= require popper
+//= require bootstrap
```
yivo: @Mons-web Same for `bootstrap-sprockets` here.

app/assets/stylesheets/_variables.scss
```diff
@@ -0,0 +1,38 @@
+// ===================================
```
yivo: @Mons-web Are this variables only for public module? If so we should keep them in personal directory (at web) so they will be isolated.

app/assets/stylesheets/general.scss
```diff
@@ -0,0 +1,10 @@
+html, body {
```
yivo: @Mons-web Could you please move it to subfolder?

app/assets/stylesheets/devise.scss
```diff
@@ -0,0 +1,27 @@
+body.devise {
```
yivo: @Mons-web Could you please move it to subfolder?

app/controllers/application_controller.rb
```diff
@@ -2,4 +2,16 @@
 
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
+
+  layout :layout_by_resource
```
yivo: @Mons-web I am sure we need to move this to `Web::ModuleController` since this also affects admin module.

yivo:
@Mons-web When you will drop fix for that two changes please squash commits into one and remove `[WIP]` from title so it can be merged by @mod.

PR #33, opened 2018-01-25
Add ability to customize design

spavlishak:


spec/views/admin/websites/show.html.erb_spec.rb
```diff
@@ -1,28 +1,23 @@
...
-    render
-    expect(rendered).to match(/Domain/)
-    expect(rendered).to match(/Title/)
-    expect(rendered).to match(/Logo/)
-    expect(rendered).to match(/Stylesheet/)
-    expect(rendered).to match(/MyText/)
-    expect(rendered).to match(/MyText/)
-    expect(rendered).to match(/Redirect Url/)
-    expect(rendered).to match(/State/)
+    account.update(role: :admin)
```
mod: You made integration testing in views testing
http://guides.rubyonrails.org/testing.html#testing-views

just move it into integration test and for view testing you can fix the previous one,
simplify them you can check only most important fields.

PR #35, opened 2018-01-25
Added relation between accounts has_one customer has_many documents

amir-budaychiev:
Added relation between accounts has_one customer has_many documents

### DIFF
--- a/db/migrate/20180124190321_create_customers.rb
+++ b/db/migrate/20180124190321_create_customers.rb
@@ -1,13 +1,16 @@
 class CreateCustomers < ActiveRecord::Migration[5.1]
   def change
     create_table :customers do |t|
-      t.string :first_name
-      t.string :last_name
-      t.date :dob
-      t.string :address
-      t.string :postcode
-      t.string :city
-      t.string :country
+      t.string      :first_name,  null: false
+      t.string      :last_name,   null: false
+      t.string      :address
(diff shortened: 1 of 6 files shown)


db/migrate/20180124190321_create_customers.rb
```diff
@@ -1,13 +1,16 @@
 class CreateCustomers < ActiveRecord::Migration[5.1]
   def change
     create_table :customers do |t|
-      t.string :first_name
-      t.string :last_name
-      t.date :dob
-      t.string :address
-      t.string :postcode
-      t.string :city
-      t.string :country
```
mod: Can you set account_id at the beginning for the table

dob means (date of birth) can you move it back below last_name.

db/migrate/20180124190951_create_documents.rb
```diff
@@ -1,14 +1,16 @@
...
+      t.string      :upload_id
+      t.string      :upload_filename
+      t.string      :upload_content_size
+      t.string      :upload_content_type
+      t.string      :doc_type
+      t.string      :doc_number
+
+      t.date        :doc_expire
+
+      t.references  :customer, null: false, index: true, foreign_key: true
```
mod: I always put my foreign keys at the beginning of the table, it actually helps natural join in SQL

PR #36, opened 2018-01-25
Add upload with refile, change forms for customers and documents

vpetrusenko:


Gemfile
```diff
@@ -13,6 +13,10 @@ gem 'cancancan',    '~> 2.0'
 gem 'bootstrap',    '~> 4.0.0.beta3'
 gem 'jquery-rails', '~> 4.3'
 gem 'kaminari',     '~> 1.1.1'
+gem 'sinatra', github: 'sinatra' #require master to get the rack >= 2.0
```
mod: why do you want to install Sinatra ?

Gemfile
```diff
@@ -13,6 +13,10 @@ gem 'cancancan',    '~> 2.0'
...
+# remove version on gem dependency to use sinatra master branch
+gem 'refile', require: 'refile/rails', github: 'pgericson/refile', branch: 'stupid-hack-sinatra-version'
```
mod: why not using the native refile ?

app/controllers/customers_controller.rb
```diff
@@ -1,58 +0,0 @@
-class CustomersController < ApplicationController
```
mod: why do you want to move the controller ?

mod:
please be careful I changed and updated the original branch

PR #50, opened 2018-01-26
Added profiles section to the admin panel

amir-budaychiev:
Added profiles section to the admin panel, added filter by state to the profiles table, added show action(displayed full info about profile and related documents)


app/controllers/admin/profiles_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Admin
+  class ProfilesController < ModuleController
+
+    def index
+      @profiles = Profile.all
+      @profiles = @profiles.order(state: params[:filter] == 'state_desc' ? :asc : :desc) if params[:filter].present?
```
mod: we don't need an order by state but a filter by state to show for so use a where

app/controllers/admin/profiles_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+      @profile    = Profile.find(params[:id])
+      @documents  = @profile.documents
+    end
+
+    def change_state
+      @profile  = Profile.find(params[:id])
+
+      return if @profile.state == params[:state]
+
+      @profile.update_column(:state, params[:state])
```
mod: this will skip validation 
please add the validator of state in the model and use save! with error handling

app/views/admin/profiles/index.html.erb
```diff
@@ -0,0 +1,40 @@
...
+      <tr>
+        <th>ID</th>
+        <th><%= link_to('State', admin_profiles_path(params[:filter] == 'state_desc' ? {filter: :state_asc} : {filter: :state_desc})) %></th>
+        <th>First name</th>
+        <th>Last name</th>
+        <th>Country</th>
+        <th colspan=3></th>
+      </tr>
+    </thead>
+    <tbody>
```
mod: you must simplify or use a helper instead of the states

app/views/admin/profiles/show.html.erb
```diff
@@ -0,0 +1,86 @@
...
+
+      <li class="list-group-item">
+        <b> State: </b>
+
+        <div class="dropdown show">
+          <a class="btn btn-secondary dropdown-toggle" href="#" role="button" id="dropdownMenuLink" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+            <%= @profile.state %>
+          </a>
+            <div class="dropdown-menu">
+              <% ['created', 'pending', 'approved', 'rejected'].each do |state|%>
```
mod: use list from Profile model

PR #55, opened 2018-01-27
Custom logo, header, footer support & helm chart fix

ashanaakh:


app/controllers/application_controller.rb
```diff
@@ -15,6 +15,36 @@ def domain_stylesheet
     return @stylesheet_url
   end
 
+  def domain_logo
+    if @logo_url.nil?
```
mod: you need to refactor my function to make a single query and get all data

ashanaakh:
@mod fixed

PR #65, opened 2018-01-28
Improvments for levels

vpetrusenko:


app/controllers/admin/profiles_controller.rb
```diff
@@ -19,6 +19,8 @@ def show
     def change_state
       @profile = Profile.find(params[:id])
       if @profile.update(state: params[:state])
+        account = Account.find(@profile.account_id)
```
mod: yes, but it doesn't verify that state == :approved 
write 

remove
```
account = Account.find(@profile.account_id)
```
use
```
@profile.account.set_level(:identity) if @profile.state == :approved
```

app/models/account.rb
```diff
@@ -18,6 +18,8 @@ def role
   end
 
   def increase_level
```
mod: make a switch case
```ruby
def set_level(step)
  case(step)
  when :mail
    self.level = 1
  when :phone
    self.level = 2
  when :identity
    self.level = 3
  when :address
    self.level = 4
  end
end
```

app/controllers/phones_controller.rb
```diff
@@ -3,7 +3,11 @@
 class PhonesController < ApplicationController
 
   def new
-    @phone = Phone.new
+    if current_account.level != 1
```
gfedorenko: not != but <. Cause person with level 2 can add another phone

vpetrusenko: oh okay, sorry

PR #69, opened 2018-01-29
Changed datepicker and sms body

gfedorenko:


app/controllers/phones_controller.rb
```diff
@@ -32,7 +32,7 @@ def verify
       session[:verif_code] = phone.generate_code
       Rails.logger.info("Sending SMS to %s with code %s" %
                         [phone.number, session[:verif_code]])
-      phone.send_sms(session[:verif_code])
+      phone.send_sms("Your verification code for Barong: #{ session[:verif_code] }")
```
calj: Please manage to get the application name from configuration ("Barong" in that case)

PR #103, opened 2018-02-06
Use JWT token for authentication

dkkoval:


config/initializers/doorkeeper.rb
```diff
@@ -106,10 +106,33 @@
   # Under some circumstances you might want to have applications auto-approved,
   # so that the user skips the authorization step.
   # For example if dealing with a trusted application.
-  # skip_authorization do |resource_owner, client|
-  #   client.superapp? or resource_owner.admin?
-  # end
+  skip_authorization do |resource_owner, client|
+    client.application.name == ENV['AUTO_AUTHORIZE_APP']
```
mod: I suggest we have a flag application.autoauth == 1

config/initializers/doorkeeper.rb
```diff
@@ -106,10 +106,33 @@
...
 
   # WWW-Authenticate Realm (default "Doorkeeper").
   # realm "Doorkeeper"
 end
+
+Doorkeeper::JWT.configure do
+  # Set the payload for the JWT token. This should contain unique information
+  # about the account.
+  token_payload do |opts|
+    p opts
```
mod: remove

config/initializers/doorkeeper.rb
```diff
@@ -106,10 +106,33 @@
...
+    account = Account.find(opts[:resource_owner_id])
+
+    {
+      email: account.email,
+      profile: account.as_json,
```
mod: reduce payload and reduce token valid time to 1h

mod: only email, level, role

config/initializers/doorkeeper.rb
```diff
@@ -106,10 +106,33 @@
...
+    }
+  end
+
+  # Set the encryption secret. This would be shared with any other applications
+  # that should be able to read the payload of the token.
+  secret_key Base64.urlsafe_decode64(Rails.application.secrets.jwt_shared_secret_key)
+
+  # Specify encryption type. Supports any algorithim in
+  # https://github.com/progrium/ruby-jwt
+  encryption_method 'RS256'
```
mod: I would suggest using:
HS256

dkkoval:
@mod yes it changes the access token

config/initializers/doorkeeper.rb
dkkoval: Then it will be better if I move it to a separated patch

PR #105, opened 2018-02-07
Unifying styles in the admin panel

amir-budaychiev:
Unifying styles in the admin panel

Fixing db:load:fake task


gfedorenko:
Add dropdown for account role editing

Issue #85, opened 2018-01-31
break-word causes mail addresses to display ugly 


yivo:
<img width="1280" alt="screen shot 2018-01-31 at 3 19 48 pm" src="https://user-images.githubusercontent.com/7421323/35625257-596f90d0-069a-11e8-82dc-07f557559121.png">

gfedorenko:
Closed by #106

PR #106, opened 2018-02-07
Minor fixes to index page

vpetrusenko:
2018-02-07-160301_1680x1050_scrot
This will appear after user achieve level 3

### DIFF
--- a/app/views/index/index.html.erb
+++ b/app/views/index/index.html.erb
@@ -21,7 +21,7 @@
             <%= link_to 'Verify your identity', new_profile_path, class: 'btn btn-primary btn-block mb-4' %>
           <% end %>
           <% if current_account.level == 3%>
-            <h2> Your identity has been successfully approved</h2>
+            <h2 class="wrap"> Your identity has been successfully approved</h2>
           <% end %>
         </div>
       </div>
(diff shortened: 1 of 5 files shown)


app/views/index/index.html.erb
```diff
@@ -8,7 +8,7 @@
 </div>
 
 <div class="row justify-content-center">
-  <div class="col-xs-12 col-sm-11 col-md-6 mt-5">
```
gfedorenko: Why did you delete this class?

vpetrusenko: @gfedorenko i remove only col-md-6 for remove between break

gfedorenko:
Investigate CSS word-break property for fixing #85

gfedorenko:
need to fix this
screenshot from 2018-02-13 14-45-01

vpetrusenko:
@gfedorenko  with last maser i see this
screencapture-localhost-3000-1518528708674

gfedorenko:
@vpetrusenko try resizing

vpetrusenko:
@gfedorenko now it's looks like this
screencapture-localhost-3000-1518598194767

PR #114, opened 2018-02-12
Return level to peatio

vpetrusenko:


app/models/account.rb
```diff
@@ -56,6 +57,11 @@ def level_set(step)
 
     save
   end
+
+  def hook_req
+    req_params = to_json(only: %i[email level])
+    RestClient.put(ENV['peatio_webhook'].to_s, req_params)
```
yssavchuk: I think it is better to use ENV.fetch['PEATIO_WEBHOOK_URL']

vpetrusenko: Rewrite according to this review

app/models/account.rb
```diff
@@ -56,6 +57,11 @@ def level_set(step)
 
     save
   end
+
+  def hook_req
```
gfedorenko: Should try to put it into level_set

vpetrusenko: @gfedorenko rewrite according to this changes

mod:
There is many things wrong about this patch,
peatio can access barong API you guys have to learn about OAuth 2.0 protocol.
https://www.youtube.com/results?search_query=oauth+2.0+protocol+explained

PR #120, opened 2018-02-13
Add JWT support to chart

ashanaakh:


ashanaakh:
@gfedorenko put label ready

PR #127, opened 2018-02-14
Update oauth callback

vpetrusenko:


db/migrate/20180109194005_devise_create_accounts.rb
```diff
@@ -34,6 +34,7 @@ def change
 
       t.string   :role, null: false, default: 'member'
       t.integer  :level, null: false, default: 0
+      t.string   :state, null: false, default: 'active'
```
gfedorenko: Don't edit previous migrations. Create new.

db/schema.rb
```diff
@@ -34,6 +34,7 @@
     t.integer "level", default: 0, null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.string "state", default: "active", null: false
```
mod: add the columns before created_at field

PR #128, opened 2018-02-14
Fix tests due to new forms

spavlishak:


gfedorenko:
@spavlishak rebase, please

PR #149, opened 2018-02-19
Add document validation

spavlishak:
Added file size, doc_number and file extension validations


gfedorenko:
@vshatravenko now images can be up to 5 megabytes

PR #153, opened 2018-02-20
Switch to grape

vpetrusenko:


app/controllers/api/v1/accounts_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'doorkeeper/grape/helpers'
+
+module API
+  module V1
+    class AccountsController < Grape::API
```
mod: rename the file and class into Accounts

app/controllers/api/v1/accounts_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      before do
+        doorkeeper_authorize!
+
+        def current_account
+          @current_account = Account.find(doorkeeper_token.resource_owner_id) if doorkeeper_token
+        end
+      end
+
+      resource :account do
```
mod: what is the meaning of resource ?

vpetrusenko: @mod instead of this we can write ' get '/account' ' , but if we want more routes with account in future better to leave this

PR #156, opened 2018-02-20
Added autoauthorized application

gfedorenko:


config/oauth_application.yml
```diff
@@ -0,0 +1,4 @@
+oauth_app_name:     'changeme'
+oauth_callback_url: 'http://changeme/callback'
```
mod: If you want to use this for seed
we shouldn't set uid and secret, for example model must generate it if empty
I bet in admin panel when we add an app we don't specify it
also we may want to load a localhost test app for env development.

about config/oauth_application.yml
I think we need a better name also we need to be able to list many applications.

config/initializers/doorkeeper.rb
```diff
@@ -106,9 +106,9 @@
   # Under some circumstances you might want to have applications auto-approved,
   # so that the user skips the authorization step.
   # For example if dealing with a trusted application.
-  # skip_authorization do |resource_owner, client|
-  #   client.superapp? or resource_owner.admin?
-  # end
+  skip_authorization do |resource_owner, client|
+    client.application.autoauth
```
mod: rename skipauth

config/application.rb
```diff
@@ -19,6 +19,7 @@ class Application < Rails::Application
     config.generators.tests  = false
     config.generators.test_framework :rspec
 
+
```
mod: remove

db/seeds.rb
```diff
@@ -12,3 +12,10 @@
 admin = Account.create(email: admin_email, password: SecureRandom.hex(20), level: 1, role: 'admin', confirmed_at: Time.now)
 
 puts 'Admin credentials: %s' % [admin.password]
+
+env_file = File.join(Rails.root, 'config', 'oauth_application.yml')
```
mod: we need to iterates over the app
I would even call it config/seed.yml

and allow other type of values likes websites.

config/oauth_application.yml
gfedorenko: @mod Valentine asked me for ability to set id and token so we don't need to redeploy peatio after we redeploy barong. This way we can just create an app with secrets that current peatio uses.

PR #158, opened 2018-02-20
Create API documentation

vpetrusenko:


gfedorenko:
Do we really need all of this dependencies and gems?

vpetrusenko:
@gfedorenko i'm not sure now, maybe we should install all dependencies except of grape-swagger with yarn

Gemfile
```diff
@@ -15,6 +15,10 @@ gem 'jquery-rails', '~> 4.3'
 gem 'kaminari',     '~> 1.1.1'
 gem 'doorkeeper',   '~> 4.2.6'
 gem 'grape'
+gem 'wine_bouncer'
```
gfedorenko: Set gem versions

vpetrusenko:
@mod okay i will fix it today and will work with swagger-ui through yaml

PR #162, opened 2018-02-21
Increase max size of upload

spavlishak:


gfedorenko:
Increased limit for Savitar. They requested 10MB

PR #163, opened 2018-02-21
Account create through API

vpetrusenko:


mod:
@vpetrusenko your patch is not finished stick to restful routes
also why you don't use Account.new validate then save.
why route is called authentification and using a find_by ?

vpetrusenko:
@mod i thought that this patch is kind of frozen, but if we need this patch i'll work on it today and fix it according to your review, most things in this patch i find in guides about creating API with devise

PR #183, opened 2018-02-23
Vault support

ashanaakh:


mod:
@ashanaakh we don't need the chart here, maybe in the kite module.
here eventually a README for installing with docker

ashanaakh:
@mod Ok

Issue #174, opened 2018-02-23
Implement new layout for barong emails


amir-budaychiev:
![Screen Shot 2018-02-19 at 7.49.51 PM.png](https://images.zenhubusercontent.com/5a902d224b5806bc2bc821a1/88ebb319-2696-43b7-9168-6265de12d540)

PR #187, opened 2018-02-24
Implement new layout for barong emails

amir-budaychiev:
Related with issue #174


app/helpers/mailer_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+  include ApplicationHelper
+
+  def determine_logo_url
+    host      = confirmation_url(Account.first) # TODO: need way to determine a host correctly, just tip to me, but it works!
+    websites  = Website.all
+    logo_url  = nil
+    websites.each do |website|
+      logo_url = website.logo if host.include?(website.domain)
+    end
+    domain_logo_tag(logo_url)
```
gfedorenko: domain_logo_tag loads white logo. So we either should load black logo or change background color to darker.

app/views/layouts/mailer.html.erb
```diff
@@ -1,13 +1,55 @@
...
+<body>
+  <table class="body-wrap">
+    <tbody>
+      <tr>
+        <td class="title-header">
+          <%= determine_logo_url %>
+        </td>
+        <td class="container">
+          <div class="content">
+            <%= yield %>
```
gfedorenko: Also we should change texts in emails a bit

amir-budaychiev:
@gfedorenko finished

app/views/layouts/mailer.html.erb
```diff
@@ -1,13 +1,55 @@
...
+          </div>
+        </td>
+      </tr>
+    </tbody></table>
```
gfedorenko: Align closing tags

app/views/layouts/mailer.html.erb
```diff
@@ -1,13 +1,55 @@
...
+</body>
```
gfedorenko: Missing </html>

app/views/devise/mailer/unlock_instructions.html.erb
```diff
@@ -1,7 +1,12 @@
-<p>Hello <%= @resource.email %>!</p>
+<p>Dear <%= @resource.email %>!</p>
+
+<p>Thank you for using with CoinField.</p>
```
gfedorenko: Replace all "CoinField" with ENV.fetch('APP_NAME',  'Barong')

mod: you can't fetch things everywhere.

gfedorenko: We can have a helper method

amir-budaychiev:
@gfedorenko recheck again please

app/mailers/application_mailer.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class ApplicationMailer < ActionMailer::Base
-  default from: 'from@example.com'
+  # default from: 'from@example.com'
```
mod: how do we define the from ?

app/helpers/mailer_helper.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module MailerHelper
+
+  def determine_logo_url
+    host      = confirmation_url(Account.first) # TODO: need way to determine a host correctly, just tip to me, but it works!
+    websites  = Website.all
```
mod: we should compare current application domain name and find it in website
we may want to make a fallback variable like APP_DOMAIN

app/views/layouts/mailer.html.erb
```diff
@@ -1,13 +1,58 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <title>Barong Mailer</title>
```
mod: We cannot write barong anywhere
and normally we need to use information from website like application name

app/views/layouts/mailer.html.erb
```diff
@@ -1,13 +1,58 @@
...
+        max-width: 600px;
+        margin: 0 auto;
+        display: block;
+      }
     </style>
   </head>
 
   <body>
-    <%= yield %>
+    <table class="body-wrap">
```
mod: are you sure table is the best way to acheive this template ?
I was thinking just a few div.

amir-budaychiev: @mod It's best practice to working with tables when we building email templates, allows us to center elements and etc. You can find a lot of interesting information on this topic, for example:
https://litmus.com/blog/the-tyranny-of-tables-why-web-and-email-design-are-so-different
https://mailbakery.com/blog/how-email-html-coding-differs-from-web-html-coding/

app/mailers/application_mailer.rb
amir-budaychiev: @mod Please, go to /config/initializers/devise.rb line 22

app/views/layouts/mailer.html.erb
mod: can we pass parameter in this way to devise emails ?
http://guides.rubyonrails.org/action_mailer_basics.html#sending-emails-with-dynamic-delivery-options

mod: @amir-budaychiev we need to pass the current hostname when the email sending was originated I doubt you understand how websites table works.

we match domain name please read our code regarding custom logo in layout.

amir-budaychiev:
@mod I found an easier way for determine hostname. Variable URL_HOST already used in production settings, so we don't have to create other variable like APP_DOMAIN. We use this in the configuration settings the mailer config.action_mailer.default_url_options.

app/helpers/mailer_helper.rb
```diff
@@ -0,0 +1,28 @@
...
+  def app_name
+    ENV.fetch('APP_NAME', 'Barong')
+  end
+
+  def url_host
+    ENV.fetch('URL_HOST', 'http://localhost:3000')
+  end
+
+  def determine_logo_url
+    websites  = Website.all
```
mod: find_by_domain

app/mailers/devise_background_mailer.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+# Mailer proxy to send devise emails in the background
+class DeviseBackgroundMailer
+
+  def self.confirmation_instructions(record, token, opts = {})
```
mitjok: This and below methods could be DRY-ed by 
```%i[confirmation_instructions reset_password_instructions ..].each { |method_name| define_singleton_method(method_name) do ..```

amir-budaychiev:
@mod Now it works good and get request.domain value and pass it via Rails cache. So, it works without background job. At the moment this is the best solution that I could found by studying the devise documentation and source code of gem devise.

PR #191, opened 2018-02-26
add swagger-ui via yarn

mitjok:


mod:
After my study can you replace this with generating the documention from a rake task
I will manage documentation portal using peatio-docs
TODO:
- [ ] install https://github.com/mermade/widdershins in yarn
- [ ] rake task for generating docs/index.md

PR #194, opened 2018-02-27
Fixed bin/setup to work in correct order

vshatravenko:


mod:
@vshatravenko we need instead to generate a seed file in the workbench

db/seeds.rb
```diff
@@ -13,11 +13,18 @@
...
 if File.exist?([Rails.root, '/config', '/seed.yml'].join)
-  seed_file = File.join(Rails.root, 'config', 'seed.yml')
-  seed = YAML.load(File.open(seed_file))
+  seed_file   = File.join(Rails.root, 'config', 'seed.yml')
+  seed        = YAML.load(File.open(seed_file))
+
+  app_seed    = seed['oauth_application']
+  application = Doorkeeper::Application.new(skipauth: app_seed['skipauth'], name: app_seed['oauth_app_name'], redirect_uri: app_seed['oauth_callback_url'])
+
+  unless ENV['WORKBENCH'].nil?
```
dkkoval: Let's do it like this: `if ENV['WORKBENCH'].in?(%w[ 1 true ])`. Also we don't need `seed.yml` in this case so check if it's the workbench before.

bin/setup
```diff
@@ -19,10 +19,13 @@ chdir APP_ROOT do
...
 
-  puts "\n=== Preparing assets ==="
-  system! 'bin/rake yarn:install assets:precompile'
-
   puts "\n== Preparing database =="
   system! 'cp config/database.yml.example config/database.yml'
   system! 'bin/rails db:create db:migrate'
+
+  puts "\n== Seeding database =="
+  system! 'bin/rails db:seed'
```
dkkoval: Let's move it to the end since it outputs some credentials and we're lazy to scroll up 😄

config/seed.yml
```diff
@@ -1,3 +1,4 @@
 oauth_application:
-  oauth_app_name:     'Client App'
-  oauth_callback_url: 'http://localhost:3001/oauth/callback'
\ No newline at end of file
+  oauth_app_name:     'Local Peatio'
```
dkkoval: Maybe simply `name`?

config/seed.yml
```diff
@@ -1,3 +1,4 @@
...
+  oauth_callback_url: 'http://localhost:8000/auth/barong/callback'
```
dkkoval: and `callback_url`

db/seeds.rb
mod: this need to move in yml files seed.yml that will be mounted.

db/seeds.rb
```diff
@@ -6,18 +6,31 @@
...
+
+  puts 'Admin credentials: %s' % [admin.password]
 
+  # Create applications from seed
+  seed['applications'].each do |app|
+    unless app['uid'].nil? && app['secret'].nil?
+      result = Doorkeeper::Application.new(skipauth: app['skipauth'], name: app['name'], redirect_uri: app['callback_url'], uid: app['uid'], secret: app['secret'])
+    else
+      result = Doorkeeper::Application.new(skipauth: app['skipauth'], name: app['name'], redirect_uri: app['callback_url'])
+    end
```
mod: why you don't simply run
Doorkeeper::Application.new(app)

vshatravenko: true :sweat_smile:

PR #198, opened 2018-02-27
Email notification when profile state is changed

spavlishak:


app/mailers/application_mailer.rb
```diff
@@ -3,4 +3,10 @@
 class ApplicationMailer < ActionMailer::Base
   default from: 'from@example.com'
   layout 'mailer'
+
+  def approved_email(profile)
+    @profile = profile
+    @account = Account.find(profile.account_id)
```
mod: normally profile.account should exist.

app/controllers/admin/profiles_controller.rb
```diff
@@ -20,7 +20,10 @@ def show
     def change_state
       @profile = Profile.find(params[:id])
       if @profile.update(state: params[:state])
-        @profile.account.level_set(:identity) if @profile.state == 'approved'
+        if @profile.state == 'approved'
+          @profile.account.level_set(:identity)
+          ApplicationMailer.approved_email(@profile).deliver
```
mod: @dkkoval @spavlishak I think we should trigger emails from model.

PR #207, opened 2018-03-01
Different gem subsets dynamic installation

mitjok:


mitjok:
```
group :plugin do
  source "https://gems.helioscloud.com" do
    plugins.each do |plugin|
      gem plugin.name, plugin.opts
    end
  end
end
```

Bundler allows to evaluate `only` ruby DSL defined in Bundler::Dsl or Bundler::Plugin::DSL. The code above can not be used.

mitjok:
@mod pls note my pull request to `bunnyware` gem

PR #227, opened 2018-03-05
Add redirect after document create

vpetrusenko:


app/controllers/index_controller.rb
```diff
@@ -5,7 +5,7 @@ def index
     redirect_to new_account_session_url unless account_signed_in?
     if current_account
       redirect_to new_phone_path if current_account.level == 1
-      redirect_to new_profile_path if current_account.level == 2
+      redirect_to new_profile_path if current_account.level == 2 && !current_account.profile
```
gfedorenko: try replacing with !current_account.profile.try(:documents)

vpetrusenko: @gfedorenko fixed

Issue #223, opened 2018-03-05
All entered data is lost when submitting form with invalid data


yivo:
![image](https://user-images.githubusercontent.com/7421323/36973161-7b39abf2-207a-11e8-874a-a195cc2edd88.png)

amir-budaychiev:
@yivo @gfedorenko Related PR #231 is merged and issue should be close.

PR #231, opened 2018-03-05
Fix all entered data which is lost when submitting form with invalid data

amir-budaychiev:
Related with issue #223


gfedorenko:
@amir-budaychiev  let's add the same for profile/new

Issue #247, opened 2018-03-06
Email input is not focused at page load, after submit password input is not focused too. Low user experience


yivo:
![image](https://user-images.githubusercontent.com/7421323/37037860-ece2f3d0-215b-11e8-8bde-c1679e631c20.png)

PR #249, opened 2018-03-07
#247 Add focus to email input at page load and password input

amir-budaychiev:
Added autofocus on page to password input


amir-budaychiev:
Related with issue #247

PR #250, opened 2018-03-07
Update the JWT Session payload

dkkoval:


config/initializers/doorkeeper.rb
```diff
@@ -118,15 +118,19 @@
...
+      exp: 4.hours.from_now.to_i,
+      sub: 'session',
+      iss: 'barong',
+      aud: opts[:scopes].all,
+
+      # unique token id (for revocation purposes)
+      jti: SecureRandom.hex(12).upcase,
+
+      account: Account.find(opts[:resource_owner_id]).as_json(
+        only: %i[uid email role level state]
```
mod: Can we have a flatten structure of account field in this payload

PR #254, opened 2018-03-12
Fix JWT payload backward compatibility

dkkoval:


mod:
we are not pushing this on master
you need to use an older version of barong and eventually merge master in it if needed.

PR #257, opened 2018-03-13
Use email as account name in Vault TOTP; allow setting issuer name in env

### LINKED ISSUE #258: Extremely difficult to understand UID, use email instead
https://github.com/rubykube/barong/blob/b826730091d82e50f75a6415bdf01b925f9c2133/lib/vault/totp.rb#L33

@dkkoval Can you please fix this ?
Also I need ability to specify issuer name with ENV.

gfedorenko:
- Pass email along with uid to vault
- Get issuer from ENV
Closes #258

### DIFF
--- a/app/controllers/security_controller.rb
+++ b/app/controllers/security_controller.rb
@@ -7,7 +7,7 @@ class SecurityController < ApplicationController
 
   # GET /security
   def enable
-    @otp = Vault::TOTP.safe_create(current_account.uid)
+    @otp = Vault::TOTP.safe_create(current_account.uid, current_account.email)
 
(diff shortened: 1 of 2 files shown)


gfedorenko:
@dkkoval yes, I tested. 2FA works fine and Google Authenticator shows email and not uid. Seems like I didn’t break it)

antonror:
selection_006

gfedorenko:
Hello @antonror 

You should configure vault. Set up Vaul server and see our 2FA docs.

antonror:
@gfedorenko thank you for such a rapid response!

PR #265, opened 2018-03-16
[WIP] Add ability to disable/enable sign up

ashanaakh:


maksim-litvinov:
@mod Do we need it or it should be closed?

PR #269, opened 2018-03-20
Changed devise logout path from delete to get request

andreymakovenko:
We need it to be able to make a sign out from frontend side.


mod:
@andreymakovenko I incorrectly create a merge commit, please remove it and rebase on master
also I don't understand why we need this,
delete session is a perfectly normal and conventional way to delete session.
who asked for this and why the other application wouldn't be able to issue a delete?
is it some kind of redirection hack?

Redirection from where?

PR #277, opened 2018-03-21
Added 2fa confirmation with otp

gfedorenko:


maksim-litvinov:
@gfedorenko I'm going to finish it

PR #283, opened 2018-03-22
Admin panel API

gfedorenko:


app/controllers/api/base.rb
```diff
@@ -1,7 +1,19 @@
 # frozen_string_literal: true
 
+require 'doorkeeper/grape/helpers'
+
 module API
   class Base < Grape::API
+    helpers Doorkeeper::Grape::Helpers
+
+      before do
+        doorkeeper_authorize!
```
andreymakovenko: not sure if we need it here, not all the endpoints need user to be authorized already
Please, make sure, if endpoint without jwt in headers will not crash like this. thanks

gfedorenko: @andreymakovenko I think, Admin should be authorized anyway in order to access admin panel.

dkkoval: crap, I forgot it's **admin** API :smile:

andreymakovenko: @gfedorenko yes, and this should go into separate admin controller, not in the base one

PR #284, opened 2018-03-22
Add ability to define title

vpetrusenko:


app/helpers/application_helper.rb
```diff
@@ -4,6 +4,15 @@
 # ApplicationHelper
 #
 module ApplicationHelper
+
+  def domain_title_tag(text)
+    if text.blank?
```
maksim-litvinov: Use inline syntax `text.blank? ? 'Barong' : text`

PR #291, opened 2018-03-26
Implemented ability to change password by API

andreymakovenko:


app/controllers/api/v1/password.rb
```diff
@@ -0,0 +1,29 @@
...
+        doorkeeper_authorize!
+
+        def current_account
+          Account.find(doorkeeper_token.resource_owner_id) if doorkeeper_token
+        end
+      end
+
+      desc 'Password related routes'
+      resource :password do
+        desc 'Change user\'s passord'
```
maksim-litvinov: password

app/controllers/api/v1/password.rb
```diff
@@ -0,0 +1,29 @@
...
+        post '/update' do
+          account = current_account
```
maksim-litvinov: Use just `put do ... end`
Use just current_account
Declare params. Don't use old_password. Doorkeeper has already verified it
Use allow_blank: false in params

Your save! will raise error and you don't handle it.
```
unless current_account.update(declared(params)
  error!(account.errors.full_messages.to_sentence)
end
```

spec/api/password_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+describe 'Password changing endpoint test' do
+  describe 'POST /api/password/update' do
+    let(:url) { '/api/password/update' }
+    let!(:password0) { 'testpassword111' }
+    let!(:password1) { 'testpassword123' }
+    let(:params0) do
+      {
+        old_password: password0,
+        new_password: password1
```
maksim-litvinov: don't use old_password, use doorkeeper shared context

andreymakovenko: it was techincal specs :)

maksim-litvinov: ok

PR #293, opened 2018-03-28
Enable logging in workbench development environment

dkkoval:


mod:
@yivo how did you solve this on peatio?

yivo:
@mod @dkkoval Just put all logging logic into single file: https://github.com/rubykube/peatio/blob/master/config/environments/shared/logger.rb

mod:
@dkkoval Please proceed in the same way of Peatio I don't want to add a WORKBENCH variable

PR #296, opened 2018-03-29
Removed target class from swagger docs

gfedorenko:


app/controllers/api/v1/base.rb
```diff
@@ -15,7 +15,6 @@ class Base < Grape::API
                                   description: 'API for barong OAuth server '
                                 },
                                 api_version: 'v1',
```
maksim-litvinov: Why we need it?

PR #297, opened 2018-03-29
Documents API

andreymakovenko:
+changed documents reference from profile to account


app/controllers/api/v1/documents.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      desc 'Account related routes'
+      resource :document do
+        desc 'Return current user documents list'
+        get '/' do
+          current_account.profile.documents.as_json(only: %i[upload doc_type doc_number doc_expire])
+        end
+
+        desc 'Return information about current resource owner'
```
mod: description is wrong

app/controllers/api/v1/documents.rb
```diff
@@ -0,0 +1,41 @@
...
+        post '/upload' do
+          profile = current_account.profile
+          return error!('Can\'t upload document, create profile first', 400) unless profile
+          document = Document.new
```
mod: Maybe Document.new(params) but make sure we filter params with a whitelist list they do in rails

app/controllers/api/v1/documents.rb
```diff
@@ -0,0 +1,41 @@
...
+          document.profile_id = profile.id
+          document.doc_type = params[:doc_type]
+          document.doc_expire = params[:doc_expire]
+          document.doc_number = params[:doc_number]
+          document.upload = params[:image]
+          return error!(document.errors.full_messages.to_sentence, 400) unless document.save
+          'Your document has been uploaded successfully'
```
mod: better retun json in a json api

db/schema.rb
```diff
@@ -44,14 +44,14 @@
...
-    t.bigint "profile_id"
     t.string "upload"
     t.string "doc_type"
     t.string "doc_number"
     t.date "doc_expire"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["profile_id"], name: "index_documents_on_profile_id"
+    t.bigint "account_id"
+    t.index ["account_id"], name: "index_documents_on_account_id"
```
mod: Do not add the field at the end, it's not a garbage bag

app/controllers/api/v1/documents.rb
```diff
@@ -0,0 +1,37 @@
...
+        end
+
+        desc 'Upload a new document for current user'
+        params do
+          requires :doc_expire, :doc_type, :doc_number, :upload
+        end
+
+        post '/' do
+          doc = Document.new(declared(params).merge(account: current_account))
+          return error!(doc.errors.full_messages.to_sentence, 400) unless doc.save
```
mod: Please write this in 5 lignes of code,
it's not about the least amount of lines
you should use a begin rescue block

is declared(params).merge(account: current_account)
what happens if the account exists
what happens if params are nil or empty?

what do you return in case of success?

Do you even know?

write me a clear structure because I have no idea

```ruby
post '/' do
  begin
    parse params
    validate params
    save params
    return success
  rescue error
     log_error
     return formatted error from helper
  end
end
```

app/controllers/api/v1/documents.rb
```diff
@@ -0,0 +1,46 @@
...
+          doc = current_account.documents.new(declared(params))
+          if doc.save
+            status 201
+          else
+            error!(doc.errors.full_messages.to_sentence, 400)
+          end
+        # temporary rescues the connection errors in fog-gooogle
+        # TODO: check workability after adding carrierwave-google-storage gem
+        rescue Excon::Error => e
+          error!('Connection error', 500)
```
maksim-litvinov: You need to add a test for that case

app/controllers/documents_controller.rb
```diff
@@ -21,7 +21,7 @@ def new
   # POST /documents
   def create
     @document = Document.new(document_params)
-    if @document.update(profile_id: current_account.profile.id)
```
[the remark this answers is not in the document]
andreymakovenko: agree

app/controllers/api/v1/documents.rb
andreymakovenko: it's very specific type of error, I will see if I can make smth

maksim-litvinov: @andreymakovenko ok, add it if possible with minimum time

app/models/document.rb
```diff
@@ -17,24 +17,24 @@ class Document < ApplicationRecord
...
 #
 #  id         :integer          not null, primary key
-#  profile_id :integer
 #  upload     :string(255)
 #  doc_type   :string(255)
 #  doc_number :string(255)
 #  doc_expire :date
 #  created_at :datetime         not null
 #  updated_at :datetime         not null
+#  account_id :integer
```
mod: position it after id and before upload field.

PR #298, opened 2018-03-29
Implemented API endpoind for receiving JWT by account and password

andreymakovenko:


app/controllers/api/v1/session.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'barong/security/access_token'
+
+module API
+  module V1
+    class Session < Grape::API
+      desc 'Session related routes'
+      resource :session do
+        post '/create' do
```
mod: POST session is (session#create)

mod: Document the params of grape

mod: http://guides.rubyonrails.org/routing.html#crud-verbs-and-actions

app/controllers/api/v1/session.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'barong/security/access_token'
+
+module API
+  module V1
+    class Session < Grape::API
+      desc 'Session related routes'
+      resource :session do
```
yivo: Should be `sessions` as per JSON API.

app/controllers/api/v1/session.rb
```diff
@@ -0,0 +1,26 @@
...
+        post '/create' do
+          acc = Account.find_by(email: params[:email])
+          return error!('401 Unauthorized', 401) unless acc
```
yivo: Makes no sense always to respond with different 401 messages. People can use it to guess what emails or apps exist. Always respond with 401 and typical message (or better no message).

spec/api/session_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'spec_helper'
```
yivo: Must be removed. See .rspec.

spec/api/session_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+describe 'Session create test' do
```
yivo: Should describe grape api.

spec/api/session_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    let!(:application) { create :doorkeeper_application }
+    subject!(:acc) do
+      create :account,
+      email: email,
+      password: password,
+      password_confirmation: password
+    end
+
+    it 'Checks if current credentials are valid and returns valid JWT' do
+      post uri, params: { email: email, password: password, application_id: application.uid }
```
yivo: Why no Content-Type is sent with headers?

spec/api/session_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    let(:check_uri) { '/api/security/renew' }
+    let!(:application) { create :doorkeeper_application }
+    subject!(:acc) do
+      create :account,
+      email: email,
+      password: password,
+      password_confirmation: password
+    end
+
+    it 'Checks if current credentials are valid and returns valid JWT' do
```
yivo: Should start from lowercase.

spec/api/session_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      post uri, params: { email: email, password: password, application_id: application.uid }
+      expect(response.status).to eq(201)
+
+      post check_uri,
+           headers: { Authorization: "Bearer #{JSON.parse(response.body)}" }
```
yivo: It is not clear what does it returns. What is `response.body`?

yivo: How does it look like and why you use JSON.parse here?

app/controllers/api/v1/session.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'barong/security/access_token'
```
yivo: This should never be required in such way in rails app. Code it in such way it is eager loaded from lib.

app/controllers/api/v1/session.rb
yivo: No need for `/create` since it is already a POST. Just `POST /sessions`.

app/controllers/api/v1/session.rb
```diff
@@ -0,0 +1,26 @@
...
+
+module API
+  module V1
+    class Session < Grape::API
+      desc 'Session related routes'
```
yivo: This description is used for the next route. Please write good one.

app/controllers/api/v1/session.rb
```diff
@@ -0,0 +1,26 @@
...
+      resource :session do
+        post '/create' do
+          acc = Account.find_by(email: params[:email])
+          return error!('401 Unauthorized', 401) unless acc
+
+          app = Doorkeeper::Application.find_by(uid: params[:application_id])
```
yivo: API lucks documentation and validation for params. Please, don't forget to validate email using grape validators.

app/controllers/api/v1/session.rb
```diff
@@ -0,0 +1,26 @@
...
+          return error!('401 Unauthorized Application', 401) unless app
+
+          if acc.valid_password? params[:password]
+            Barong::Security::AccessToken.create params[:expires_in], acc.id, app
```
yivo: Is `expires_in` validated? Type? Limitations?

app/controllers/api/v1/session.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module API
+  module V1
+    class Session < Grape::API
+      desc 'Session related routes'
+      resource :sessions do
+        desc 'Start a new session'
+        params do
+          requires :email
```
maksim-litvinov: add `allow_blank: false` for all required params

andreymakovenko: agree

PR #316, opened 2018-04-03
Document upload hotfix

andreymakovenko:


config/initializers/carrierwave.rb
```diff
@@ -5,15 +5,14 @@
 require 'carrierwave/storage/fog'
 
 CarrierWave.configure do |config|
-  if 'Google'.casecmp(Rails.application.secrets.storage_provider) == 0
-    config.fog_provider = 'fog/google'
+  if Rails.application.secrets.storage_provider.eql? 'Google'
```
ashanaakh: Why you did that?

config/initializers/carrierwave.rb
```diff
@@ -5,15 +5,14 @@
...
     config.fog_credentials = {
       provider: 'Google',
       google_storage_access_key_id: Rails.application.secrets.storage_access_key,
       google_storage_secret_access_key: Rails.application.secrets.storage_secret_key
     }
     config.fog_directory = Rails.application.secrets.storage_bucket_name
-  elsif 'AWS'.casecmp(Rails.application.secrets.storage_provider) == 0
+  elsif Rails.application.secrets.storage_provider.eql? 'AWS'
```
ashanaakh: Better to left case insensitive compare.
*Example:*
 Some customers can use `Aws` or `aws` instead of `AWS`.

Issue #306, opened 2018-04-02
Add API for manage labels


maksim-litvinov:
We need CRUD actions to manage labels:
API must follow REST.

Label is a resource with next fields:
- key: `string`
- value: `string`
- scope: `['public', 'private']`

Example of labels:

- email: verified
- color: blue
- size: 42

#### Public scope

Tags with public scope can be created from API
and update they are 100% owned by the user.

#### Private scope

Private tags can be only created from Barong admin panel
they are READ ONLY from api.

mitjok:
@maksim-litvinov could you reflect this slack post of @mod in this ticket
https://rubykube.slack.com/archives/C8XSU8WJW/p1522426201000507

maksim-litvinov:
@mitjok I updated scopes

maksim-litvinov:
Merged to master

PR #328, opened 2018-04-04
#306: labels api

mitjok:


app/models/label.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+# Resposible for storing configurations
+class Label < ApplicationRecord
+  belongs_to :account
+
+  enum scope: { general: 0, restricted: 1 }
```
maksim-litvinov: Where is the third scope?
According to slack message it should be 3 scopes: private, application, client.
In your case restricted is a private, but what is a general?

@mod Can you clarify required scopes?

app/models/label.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  validates :scope,
+            inclusion: { in: scopes.keys }
+
+  validates :key,
+            length: 3..255,
+            format: { with: /\A[a-z0-9_-]+\z/ },
+            uniqueness: { scope: :account_id }
+
+  validates :value,
```
maksim-litvinov: We need format validation here `format: { with: /\A[A-Za-z0-9_-]+\z/ }`

spec/api/labels_spec.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'Labels API.' do
+  let!(:account) { create :account }
+  let!(:label) { create :label, account: account }
+
+  let!(:access_token) { create :doorkeeper_token, resource_owner_id: account.id }
```
maksim-litvinov: you should use doorkeeper shared context here.
Example https://github.com/rubykube/barong/blob/87aa2a2690236eb71ed2c48ced48fe02f0878333/spec/api/v1/accounts_spec.rb#L4-L7

app/controllers/api/v1/labels.rb
```diff
@@ -0,0 +1,79 @@
...
+
+        desc 'Update a label.'
+        params do
+          requires :key, type: String, allow_blank: false, desc: 'Label key.'
+          requires :value, type: String, allow_blank: false, desc: 'Label value.'
+        end
+        patch ':key' do
+          current_account.labels.find_by!(key: params[:key]).tap do |label|
+            if label.general?
+              label.update(value: params[:value])
```
maksim-litvinov: @mod Do we need delete action?

db/migrate/20180402133658_create_labels.rb
```diff
@@ -0,0 +1,13 @@
+class CreateLabels < ActiveRecord::Migration[5.1]
+  def change
+    create_table :labels do |t|
+      t.string :key, null: false
+      t.string :value, null: false
+      t.integer :scope, null: false, default: 0
+      t.references :account
```
maksim-litvinov: add foreign_key for account

app/models/label.rb
mitjok: @maksim-litvinov #306 was updated by @mod recently. Two scopes left.

app/models/label.rb
mitjok: thx, done

spec/api/labels_spec.rb
mitjok: thx, done

db/migrate/20180402133658_create_labels.rb
mitjok: thx, done

Gemfile.lock
```diff
@@ -371,7 +378,9 @@ DEPENDENCIES
   fog-google (~> 0.1.0)
   fontello_rails_converter
   grape (~> 1.0)
+  grape-entity (~> 0.7.1)
   grape-swagger (~> 0.28)
+  grape_on_rails_routes (~> 0.3.2)
```
mod: is it absolutely necessary can we do it without this rails_routes gem ?

app/controllers/api/v1/labels.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      resource :labels do
+        desc 'List all labels for current account.'
+        get do
+          present current_account.labels, with: API::Entities::Label
+        end
+
+        desc 'Return a label by key.'
+        params do
+          requires :key, type: String, allow_blank: false, desc: 'Label key.'
```
mod: This is a security hole, aren't you missing current_account
you can only see your labels.

app/controllers/api/v1/labels.rb
```diff
@@ -0,0 +1,90 @@
...
+          requires :value, type: String, allow_blank: false, desc: 'Label value.'
+        end
+        post do
+          label =
+            current_account.labels.new(
+              key: params[:key],
+              value: params[:value],
+              scope: 'general'
```
mod: there is scope private or public no general scope.

app/controllers/api/entities/label.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module API
+  module Entities
+    class Label < Grape::Entity
+      format_with(:iso_timestamp, &:iso8601)
+
+      expose :key, documentation: { type: 'String', desc: 'Label key. [a-z0-9_-]+ should be used. Min - 3, max - 255 characters.' }
+      expose :value, documentation: { type: 'String', desc: 'Label value. Min - 3, max - 255 characters.' }
+      expose :scope, documentation: { type: 'String', desc: "Label scope: 'general' or 'restricted'" }
+      expose :account_id,
```
mod: could he avoid returning account_id in this entity?

app/controllers/api/v1/labels.rb
```diff
@@ -0,0 +1,90 @@
...
+        desc 'Update a label.'
+        params do
+          requires :key, type: String, allow_blank: false, desc: 'Label key.'
+          requires :value, type: String, allow_blank: false, desc: 'Label value.'
+        end
+        patch ':key' do
+          label = current_account.labels.find_by(key: params[:key])
+          return error!('Couldn\'t find Label', 404) if label.blank?
+
+          if label.general?
```
mod: I would block those routes at an earlier stage we would not authorize Updates / Delete for scope private.

app/controllers/api/v1/labels.rb
```diff
@@ -0,0 +1,90 @@
...
+            label.errors.add(:base, 'Can\'t update Label.')
+            error!(label.errors.as_json(full_messages: true), 422)
+          end
+        end
+
+        desc 'Delete a label.'
+        params do
+          requires :key, type: String, allow_blank: false, desc: 'Label key.'
+        end
+        delete ':key' do
```
mod: delete must be also subject to scope.

app/models/account.rb
```diff
@@ -12,6 +12,7 @@ class Account < ApplicationRecord
 
   has_one :profile, dependent: :destroy
   has_many :phones, dependent: :destroy
+  has_many :labels
```
mod: we can make dependent: :destroy

app/models/label.rb
mod: scope is 'public' or 'private'

db/schema.rb
```diff
@@ -54,6 +54,17 @@
     t.index ["profile_id"], name: "index_documents_on_profile_id"
   end
 
+  create_table "labels", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
+    t.string "key", null: false
+    t.string "value", null: false
+    t.integer "scope", default: 0, null: false
```
mod: use a string for scope I am planning on having more values

db/schema.rb
```diff
@@ -54,6 +54,17 @@
...
+    t.bigint "account_id"
```
mod: put account_id at the beginning of the table.

db/schema.rb
maksim-litvinov: @mod it is enum. We can add any enums. ActiveRecord show us a string but saves it as integer

app/models/label.rb
mitjok: @mod @maksim-litvinov unfortunately `public`, `private` is reserved words, so I replaced them with synonyms `general`, `restricted`

Gemfile.lock
mitjok: @mod, @maksim-litvinov I think it could be moved to `development` section on Gemfile

app/controllers/api/v1/labels.rb
mitjok: @mod, do `restricted` (private) labels always has an account assigned? I wasn't sure so made a general assumption. If `restricted` always has account assigned, then this will be fixed

app/controllers/api/v1/labels.rb
mitjok: @mod @maksim-litvinov unfortunately public, private is reserved words, so I replaced them with synonyms general, restricted

app/controllers/api/entities/label.rb
mitjok: sure

app/models/account.rb
mitjok: @mod This is done on db level by foreign key constraint `on_delete: :cascade`. No need to call AR callbacks (`dependent: :destroy`)

app/controllers/api/v1/labels.rb
mitjok: @mod, thx

db/schema.rb
mitjok: @mod sure

Gemfile.lock
mitjok: moved

app/controllers/api/v1/labels.rb
mitjok: fixed

app/controllers/api/entities/label.rb
mitjok: done

app/controllers/api/v1/labels.rb
mitjok: `before` blocks don't understand filtering by action name, so done this in the action's code

db/schema.rb
mitjok: done

app/controllers/api/v1/labels.rb
```diff
@@ -0,0 +1,86 @@
...
+        end
+
+        desc "Delete a label  with 'general' scope."
+        params do
+          requires :key, type: String, allow_blank: false, desc: 'Label key.'
+        end
+        delete ':key' do
+          label = current_account.labels.find_by(key: params[:key])
+          return error!('Couldn\'t find Label', 404) if label.blank?
+          return error!('Can\'t update Label.', 400) if label.private?
```
mod: better use 401

app/controllers/api/v1/labels.rb
```diff
@@ -0,0 +1,86 @@
...
+              scope: 'public'
+            )
+          if label.save
+            present label, with: API::Entities::Label
+          else
+            error!(label.errors.as_json(full_messages: true), 422)
+          end
+        end
+
+        desc "Update a label with 'general' scope."
```
mod: make sure we don't use the word 'general'

PR #327, opened 2018-04-04
Feature/2fa api

maksim-litvinov:
Closes #300 
Fixes #237

PR #338, opened 2018-04-07
Bugfix/2fa enable after verify

maksim-litvinov:


app/controllers/api/v1/security.rb
```diff
@@ -11,21 +11,79 @@ class Security < Grape::API
...
 
         def current_application
           doorkeeper_token.application
         end
+
+        def with_vault_error_handler
+          yield
+        rescue Vault::HTTPClientError => e
+          Rails.logger.error "#{e.class}: #{e.message}"
+          error!(e.message, 422)
```
andreymakovenko: I think what status should be 500 here

app/controllers/sessions_controller.rb
```diff
@@ -22,22 +22,19 @@ def confirm
...
   def otp_enabled?
-    uid = find_uid_by_params_email
-    return false unless uid.present?
-    Vault::TOTP.exist?(uid)
+    current_account&.otp_enabled
   end
 
   def otp_verify
-    return if Vault::TOTP.validate?(find_uid_by_params_email, params[:otp])
+    return if Vault::TOTP.validate?(current_account&.uid, params[:otp])
```
andreymakovenko: typo?

maksim-litvinov:
merged to #327

PR #386, opened 2018-04-20
Implement rejected UI message and state change on document upload

andreymakovenko:


app/controllers/index_controller.rb
```diff
@@ -5,6 +5,6 @@ def index
     return redirect_to new_account_session_url unless account_signed_in?
 
     redirect_to new_phone_path if current_account.level == 1
-    redirect_to new_profile_path if current_account.level == 2 && current_account.documents.blank?
+    redirect_to new_profile_path if current_account.level == 2 && current_account.documents.nil?
```
maksim-litvinov: `redirect_to new_profile_path if false` ???

app/views/index/index.html.erb
```diff
@@ -7,11 +7,12 @@
         <div class="d-flex justify-content-center">
           <% if current_account.level == 1 %>
             <%= link_to 'Verify your phone', new_phone_path, class: 'btn btn-primary btn-block mb-4' %>
-          <% end %>
-          <% if current_account.level == 3%>
+          <% elsif current_account.level == 3 && current_account.profile.state == 'approved' %>
```
maksim-litvinov: should be in model

app/views/index/index.html.erb
```diff
@@ -7,11 +7,12 @@
...
             <h2 class="wrap"> Your identity has been successfully approved</h2>
-          <% else %>
+          <% elsif current_account.level == 2 && current_account.profile.state == 'pending' %>
```
maksim-litvinov: should be in model

PR #411, opened 2018-04-26
Implemented account level changing logic with labels

andreymakovenko:


app/models/label.rb
```diff
@@ -26,6 +26,23 @@ class Label < ApplicationRecord
...
             length: 3..255,
             format: { with: /\A[A-Za-z0-9_-]+\z/ }
+
+  after_commit :change_level_if_label_mapped, on: [:create, :update]
+  after_destroy :change_level_if_mapped_label_destroyed
+
+private
+
+  def change_level_if_label_mapped
+    level = Level.find_by(key: key)
```
maksim-litvinov: why 3 duplicates lines?

spec/models/label_test.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+RSpec.describe Label, type: :model do
+  it { should belong_to(:account) }
+
+  describe 'update account level if label defined as level' do
+    subject(:account) { create :account }
+    let(:level_1) { Level.create(id: 1, key: 'email',    value: 'verified', description: 'User has confirmed his email')  }
+    let(:level_2) { Level.create(id: 2, key: 'phone',    value: 'verified', description: 'User has confirmed his phone')  }
```
maksim-litvinov: use factory bot

maksim-litvinov: use explicit name `email_verified_level` istead of `level_1`

spec/models/label_test.rb
```diff
@@ -0,0 +1,28 @@
...
+    let(:level_3) { Level.create(id: 3, key: 'identity', value: 'verified', description: 'User is not in sanction list') }
+    let(:level_4) { Level.create(id: 4, key: 'document', value: 'approved', description: 'User document is approved')   }
+
+    it 'changes account level consistently' do
+      expect(account.level).to eq(0)
+      Label.create(account: account, key: level_1.key, value: level_1.value)
```
maksim-litvinov: add more specs. It should not be just one spec. I think you miss some steps

PR #435, opened 2018-04-30
Remove state from profile

### LINKED ISSUE #442: Public labels can prevent creation of private labels with the same key value.
In case if user account already have a label with crucial key, like phone, the verification of a first phone or a new one will lead to an error, due to the already existing label. 
This is preventing adding numerous of phone numbers to the account.

### LINKED ISSUE #443: Rewrite mailers due to state removal
After deleting states mailer stopped to work

### LINKED ISSUE #445: Logic error connected with changes of critical label Key
When you will add some important labels like key:email value:verified scope:private, you will get lvl 1.
But if you change this label key to something else, like "barong", the level will not be changed.
selection_210
selection_211
selection_212
selection_213

m-an:
closes #445 
closes #443 
closes #442

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -6,7 +6,8 @@
 class Document < ApplicationRecord
   mount_uploader :upload, UploadUploader
 
-  TYPES = ['Passport', 'Identity card', 'Driver license'].freeze
+  TYPES = %w[Passport Identity card Driver license].freeze
+  STATES = %w[verified pending rejected].freeze
 
   belongs_to :account
   validates :doc_type, :doc_number, :doc_expire, :upload, presence: true
@@ -17,6 +18,18 @@ class Document < ApplicationRecord
   validates_format_of :doc_expire,
                       with: /\d{4}\-\d{2}\-\d{2}/,
                       message: 'Date must be in the following format: yyyy-mm-dd'
+  after_commit :create_or_update_document_label, on: :create
+
+private
+
+  def create_or_update_document_label
+    account_document_label = account.labels.find_by(key: :document)
+    if account_document_label.nil?
+      account.labels.create(key: :document, value: :pending, scope: :private)
+    elsif account_document_label.value != 'verified'
+      account_document_label.update(value: :pending)
+    end
+  end
 end
 
 # == Schema Information
--- a/app/models/label.rb
+++ b/app/models/label.rb
@@ -39,6 +39,15 @@ private
   def update_level_if_label_defined
     return unless Level.exists?(key: key)
     account.reload.update_level
+    send_document_review_notification if key == 'document'
+  end
+
+  def send_document_review_notification
+    if value == 'verified'
+      ProfileReviewMailer.approved(account).deliver_now
+    elsif value == 'rejected'
+      ProfileReviewMailer.rejected(account).deliver_now
+    end
   end
 end
 
--- a/app/models/label.rb
+++ b/app/models/label.rb
@@ -25,7 +25,7 @@ class Label < ApplicationRecord
   validates :key,
             length: 3..255,
             format: { with: /\A[A-Za-z0-9_-]+\z/ },
-            uniqueness: { scope: :account_id }
+            uniqueness: { scope: [:account_id, :scope] }
 
   validates :value,
             length: 3..255,
@@ -37,7 +37,7 @@ class Label < ApplicationRecord
 private
 
   def update_level_if_label_defined
-    return unless Level.exists?(key: key)
+    return unless scope == 'private'
     account.reload.update_level
     send_document_review_notification if key == 'document'
   end
@@ -45,14 +45,16 @@ private
   def send_document_review_notification
     if value == 'verified'
       ProfileReviewMailer.approved(account).deliver_now
+      EventAPI.notify('system.document.verified', account_uid: account.uid)
     elsif value == 'rejected'
       ProfileReviewMailer.rejected(account).deliver_now
+      EventAPI.notify('system.document.rejected', account_uid: account.uid)
     end
   end
 end
 
 # == Schema Information
-# Schema version: 20180402133658
+# Schema version: 20180507162420
 #
 # Table name: labels
 #
@@ -67,7 +69,7 @@ end
 # Indexes
 #
 #  index_labels_on_account_id                    (account_id)
-#  index_labels_on_key_and_value_and_account_id  (key,value,account_id) UNIQUE
+#  index_labels_on_key_and_scope_and_account_id  (key,scope,account_id) UNIQUE
 #
 # Foreign Keys
 #
--- a/app/models/phone.rb
+++ b/app/models/phone.rb
@@ -4,7 +4,7 @@
 # Class Phone
 #
 class Phone < ApplicationRecord
-  belongs_to :account, optional: true
+  belongs_to :account
 
   validates :number, phone: true
 
--- a/app/models/phone.rb
+++ b/app/models/phone.rb
@@ -4,7 +4,7 @@
 # Class Phone
 #
 class Phone < ApplicationRecord
-  belongs_to :account
+  belongs_to :account, optional: true
 
   validates :number, phone: true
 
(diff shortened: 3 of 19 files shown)


app/models/document.rb
```diff
@@ -17,6 +18,18 @@ class Document < ApplicationRecord
   validates_format_of :doc_expire,
                       with: /\d{4}\-\d{2}\-\d{2}/,
                       message: 'Date must be in the following format: yyyy-mm-dd'
+  after_commit :create_or_update_document_label, on: :create
+
+private
+
+  def create_or_update_document_label
```
maksim-litvinov: I can't understand why do you duplicate label logic in model and controller?

app/models/label.rb
```diff
@@ -37,13 +37,24 @@ class Label < ApplicationRecord
...
-    return unless Level.exists?(key: key)
+    return unless scope == 'private'
     account.reload.update_level
+    send_document_review_notification if key == 'document'
+  end
+
+  def send_document_review_notification
+    if value == 'verified'
+      ProfileReviewMailer.approved(account).deliver_now
+      EventAPI.notify('system.document.verified', account_uid: account.uid)
```
maksim-litvinov: add `TODO: Need to update specs/event_api.md` if you don't add it in current request. Specs are important for users of Events API

app/models/phone.rb
```diff
@@ -4,7 +4,7 @@
 # Class Phone
 #
 class Phone < ApplicationRecord
-  belongs_to :account
+  belongs_to :account, optional: true
```
maksim-litvinov: Why optional?

app/views/admin/profiles/show.html.erb
```diff
@@ -37,14 +37,14 @@
     </li>
 
     <li class="list-group-item">
-      <b> State: </b>
+      <b> Document: </b>
       <div class="dropdown show">
         <a class="btn btn-secondary dropdown-toggle" href="#" role="button" id="dropdownMenuLink" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-          <%= @profile.state.upcase_first %>
+          <%= @profile.account.labels.find_by(key: 'document', scope: 'private')&.value %>
```
maksim-litvinov: Never do any ActiveRecord `find, where` in a view

app/views/admin/profiles/show.html.erb
```diff
@@ -37,14 +37,14 @@
...
         </a>
         <div class="dropdown-menu">
-          <% Profile::STATES.each do |state|%>
-            <%= link_to(state.upcase_first, change_state_admin_profile_path(@profile, state: state), method: :put, class: 'dropdown-item') unless @profile.state == state%>
+          <% Document::STATES.each do |state|%>
+            <%= link_to(state.upcase_first, document_label_admin_profile_path(@profile, state: state), method: :put, class: 'dropdown-item') unless @profile.account.labels.find_by(key: :document).value == state%>
```
maksim-litvinov: same as above. And use short string. It's hard to read

app/views/profile_review_mailer/approved.html.erb
```diff
@@ -1,7 +1,11 @@
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
 <body>
   <div>
-    <p> Hello, <%= @profile.first_name %> <%= @profile.last_name %>!</p>
+    <% if @profile.nil? %>
+      <p> Hello! </p>
+    <% else %>
+      <p> Hello, <%= @profile.first_name %> <%= @profile.last_name %>!</p>
```
maksim-litvinov: We can use just `@profile.name` here
```
class Profile
  def name
    "#{first_name} #{last_name}"
  end
end
```

app/views/profile_review_mailer/rejected.html.erb
```diff
@@ -1,7 +1,11 @@
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
 <body>
   <div>
-    <p> Hello, <%= @profile.first_name %> <%= @profile.last_name %>!</p>
+    <% if @profile.nil? %>
+      <p> Hello! </p>
+    <% else %>
+      <p> Hello, <%= @profile.first_name %> <%= @profile.last_name %>!</p>
```
maksim-litvinov: same as above

spec/models/document_spec.rb
```diff
@@ -5,4 +5,63 @@
...
   it { should belong_to(:account) }
+
+  context 'Document creation' do
+    let!(:current_account) { create(:account) }
+    let(:create_document) { create :document, account: current_account }
+
+    context 'when it is first document' do
+      it 'adds new document label' do
+        expect(current_account.labels).to be_empty
+        create_document
```
maksim-litvinov: Suggest to do `expect { create_document }.to change { current_account.labels.count }.from(0).to(1)`
probably some `reload` needed

spec/models/document_spec.rb
```diff
@@ -5,4 +5,63 @@
...
+        expect(current_account.labels).to_not be_empty
+      end
+
+      it 'new document label is document: pending' do
+        create_document
+        expect(current_account.labels.first.key).to eq 'document'
```
maksim-litvinov:
```
 let(:document_label) { current_account.labels.first }
...
expect(document_label.key).to eq 'document'
```

spec/models/document_spec.rb
```diff
@@ -5,4 +5,63 @@
...
+        create :label,
+               scope: 'private',
+               key: 'document',
+               value: 'rejected',
+               account: current_account
+      end
+
+      it 'does not add new label' do
+        create_document
+        expect(current_account.labels.count).to eq 1
```
maksim-litvinov: how do I know that next is true "does not add new label".
probably it should be expect { create_document }.to_not change { Label.count }

spec/models/document_spec.rb
```diff
@@ -5,4 +5,63 @@
...
+        create :label,
+               scope: 'private',
+               key: 'document',
+               value: 'verified',
+               account: current_account
+      end
+
+      it 'does not add new label' do
+        create_document
+        expect(current_account.labels.count).to eq 1
```
maksim-litvinov: same as above

spec/models/document_spec.rb
```diff
@@ -5,4 +5,63 @@
...
+      end
+
+      it 'remains value verified' do
+        create_document
+        expect(current_account.labels.first.value).to eq 'verified'
```
maksim-litvinov: same as above

spec/models/label_spec.rb
```diff
@@ -89,5 +89,115 @@
         expect(account.reload.level).to eq 0
       end
     end
+
+    context 'document label changes' do
+      let!(:account) { create(:account) }
+      let(:mailer_deliveries) { ActionMailer::Base.deliveries }
```
maksim-litvinov: why not `let(:last_mailer_delivery) { ActionMailer::Base.deliveries.last }`

spec/models/label_spec.rb
```diff
@@ -89,5 +89,115 @@
...
+        create :label,
+               account: account,
+               key: email_verified_level.key,
+               value: email_verified_level.value,
+               scope: 'private'
+      end
+
+      context 'can be created with different scopes' do
+        it 'account has both labels' do
+          expect(account.labels.first).to eq label_public
```
maksim-litvinov: `expect(account.labels).to eq [label_public, label_private]`

spec/models/label_spec.rb
```diff
@@ -89,5 +89,115 @@
...
+
+        context 'when public label changes' do
+          it 'account level does not change' do
+            expect { label_public.update(value: 'rejected') }.to_not change { account.reload.level }
+          end
+        end
+      end
+    end
+
+    context 'with public labels duplicating levels' do
```
maksim-litvinov: Wrong name. It's not implicit reason for spec.
I suggest `when account has all required label for level 4 but in public scope`

tamadamas (changes requested):

tamadamas (approved):

maksim-litvinov:
already merged with #460 branch

PR #439, opened 2018-05-02
Use jwt instead of doorkeeper access token in api sessions

### LINKED ISSUE #433: Use jwt for API instead of Doorkeeper Access Token

maksim-litvinov:
Closes #433


maksim-litvinov:
need to test

PR #441, opened 2018-05-03
Add api_key resource and generate jwt session

### LINKED ISSUE #211: Ability to manage trading tokens from barong
As a user I want to access a barong api where I can create, delete tokens and configure api tokens.

call it /api/v1/keypairs
CRUD

user can create read update delete

We need a page for example security/keys
in which you can create a new api key
form is

## Schema

- name
- key: ~20 chars
- secret: ~40 chars
- scopes: space separated scopes (strings) oauth scopes "peatio:trade"
- lifetime: number of second each JWT will live for, between 10 sec and 7200 sec
- timestamps

```python
import hmac
import hashlib

message = nonce + customer_id + api_key
signature = hmac.new(
    API_SECRET,
    msg=message,
    digestmod=hashlib.sha256
).hexdigest().upper()
```

Additional feature for later
screen shot 2018-04-18 at 17 47 28

maksim-litvinov:
Closes #211


app/api/entities/api_key.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Entities
+  class ApiKey < Grape::Entity
```
yivo: @maksim-litvinov Hey. We should **always** follow convention for naming. API is acronym so it cant be named like Api or ApI.

maksim-litvinov:
Need to add 2fa check on API Key CRUD and update specs

PR #446, opened 2018-05-04
Send emails if they are enabled. Add SKIP_EMAILS env

maksim-litvinov:


nchigwamba:
@mod What is the purpose of this change?

PR #447, opened 2018-05-07
Added metadata field for document

gfedorenko:


app/api/user_api/v1/documents.rb
```diff
@@ -15,6 +15,7 @@ class Documents < Grape::API
         desc 'Upload a new document for current user'
         params do
           requires :doc_expire, :doc_type, :doc_number, :upload
+          optional :metadata, type: Hash, desc: 'Any key:value pairs'
```
maksim-litvinov: you need to add api specs for new field

spec/api/user_api/v1/documents_spec.rb
```diff
@@ -19,11 +19,27 @@
...
+
     it 'Checks if params are ok and returns success' do
       post '/api/v1/documents', headers: headers, params: params
       expect(response.status).to eq(201)
     end
 
+    it 'Creates document with optional params and returns success' do
+      post '/api/v1/documents', headers: headers, params: params.merge(optional_params)
+      expect(response.status).to eq(201)
+      document = Document.last
```
andreymakovenko: we don't create variables in 'it' blocks
use `let()` instead
`let(:last_document) { Document.last }`
but I recommend to use
`let(:account_last_document) { account.documents.last }`

spec/api/user_api/v1/documents_spec.rb
```diff
@@ -19,11 +19,27 @@
...
+        }
+      }
+    end
+
     it 'Checks if params are ok and returns success' do
       post '/api/v1/documents', headers: headers, params: params
       expect(response.status).to eq(201)
     end
 
+    it 'Creates document with optional params and returns success' do
```
andreymakovenko: Please, add a new `describe` block under your `it` block
Add into it:
`let(:do_request) { post '/api/v1/documents', headers: headers, params: params.merge(optional_params) }`
and use inside your `it` blocks like expect { do_request }.to change { account.reload.documents.count }

PR #455, opened 2018-05-10
Added password strength validation

gfedorenko:


maksim-litvinov:
Hm, I'm not sure it is more secure than previous passwords

PR #477, opened 2018-05-17
Add validations for profile, add datepicker and country select for profile admin panel

maksim-litvinov:


pkucherenk0:
It is possible to type "space+space+ latter" in the First name, Last name and City and this input will be valid. Please add validation to spaces in the beginning and in the end of a line, so it will be possible to type "George+space+Bernard".
Also it will be nice to have some validations in Address (allow only space,comma,dot,numbers and alphabets).
For the postal code should consider to add letters and space, here are the examples of different postal codes:
selection_291
Also some cosmetics:
selection_292

maksim-litvinov:
Approved by qa testing

mod:
we need a rebase

PR #478, opened 2018-05-17
Update event specs

maksim-litvinov:


maksim-litvinov:
Will be merge as part of that request https://github.com/rubykube/barong/pull/482

PR #481, opened 2018-05-17
Fix labels bug with changing scope

### LINKED ISSUE #476: Changing already created labels scope from private to public is not changing the level of users account.
STR:
 
1. Go to the Barong admin panel.
2. Find account with labels in scope private and level that goes with it.
Example:
selection_245
3. Click on the email address to get to user profile.
4. Change labels with private scope to public.

Result: The level of user was not changed .
selection_244
Expected: The level of user should decrees according to current existing private labels.

andreymakovenko:
closes #476

### DIFF
--- a/app/models/label.rb
+++ b/app/models/label.rb
@@ -39,7 +39,7 @@ class Label < ApplicationRecord
 private
 
   def update_level_if_label_defined
-    return unless scope == 'private'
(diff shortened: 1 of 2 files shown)


mod:
@andreymakovenko I don't understand the fix,
if label changes we need to recalculate, why changing this
why do you want to recalculate if scope change?

if scope changes from public to private it should recalculate

PR #483, opened 2018-05-21
Add ability to configure lockable

maksim-litvinov:
Added next ENV variables

`ENABLE_LOCKABLE: ['true', 'false']`
`UNLOCK_STRATEGY: 'both' by default`
 Defines which strategy will be used to unlock an account.
  * 'email' = Sends an unlock link to the user email
  * 'time'  = Re-enables login after a certain amount of time (see :unlock_in below)
  * 'both'  = Enables both strategies
  * 'none'  = No unlock strategy. You should handle unlocking by yourself.

`MAXIMUM_ATTEMPTS: 5 by default`
 Number of authentication tries before locking an account if lock_strategy is failed attempts.

`UNLOCK_INTERVAL (in seconds) 3600 by default`
 Time interval to unlock the account if 'time' or 'both' is enabled as unlock_strategy.

`LAST_ATTEMPT_WARNING: ['true', 'false']`
 Warn on the last attempt before the account is locked.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,6 @@ cache:
   - bundler
   - yarn
 
-branches:
-  only:
-    - "master"
-
 env:
   - RAILS_ENV=test
     RAKE_ENV=test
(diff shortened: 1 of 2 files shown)


mod:
@maksim-litvinov @mkuvshynnikov For who is this feature ?
why do we make it an option ?
can't we set a good default behavior instead?

maksim-litvinov:
@mod BitPesa said they want to disable lockable for their platform and they want ability to configure

maksim-litvinov:
Approved by qa testing

maksim-litvinov:
We will implement it another way without devise lockable

PR #484, opened 2018-05-22
Check required environments on barong starting

### LINKED ISSUE #457: Check all required ENV in production environment and raise for missing vars
Code fails without required env variables.
We need to check env variables on first deploy or update new version application to ensure everything settled properly.

maksim-litvinov:
Closed #457


mod:
good idea, but we have a conflict from last merge

PR #494, opened 2018-05-25
Added ability to define levels by env variable

andreymakovenko:
Changed level setting way in account model from level id to level order index to be able to redefine levels.
Removed description from mandatory fields for easier level definition in ENV.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -48,9 +48,9 @@ class Account < ApplicationRecord
                  .map { |l| [l.key, l.value].join ':' }
     levels = Level.all.order(id: :asc)
 
-    levels.each do |lvl|
+    levels.each_with_index do |lvl, index|
       break unless tags.include?(lvl.key + ':' + lvl.value)
-      account_level = lvl.id
+      account_level = index + 1
     end
 
     update(level: account_level)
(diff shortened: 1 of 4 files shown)


app/models/account.rb
```diff
@@ -48,9 +48,9 @@ def update_level
                  .map { |l| [l.key, l.value].join ':' }
     levels = Level.all.order(id: :asc)
 
-    levels.each do |lvl|
+    levels.each_with_index do |lvl, index|
       break unless tags.include?(lvl.key + ':' + lvl.value)
-      account_level = lvl.id
+      account_level = index + 1
```
mod: to me lvl.id is the level, and your change is using the array index
why would you like to use the array index ?

db/seeds.rb
```diff
@@ -82,13 +82,24 @@
   result[:applications].push(app.as_json(only: %i[name redirect_uri uid secret]))
 end
 
+# Example env: 'phone:verified,email:verifed,document:verified,identity:verified'
 logger.info '---'
 logger.info 'Seeding levels'
-seeds['levels'].each do |level_data|
-  next if Level.exists?(level_data)
-  Level.create!(level_data)
+levels = ENV['LEVELS']
```
mod: we don't need environment variable
you edit level by seed file

app/models/account.rb
andreymakovenko: to be able to edit levels in future:
index is actually an order id of levels database, so this way, for example, you will be able to delete level 3(if you suddenly need to) and what was level 4 will become level 3(by order, not by id), or you will be able even to reinitiolize all the levels by deleting them all and creating new ones

PR #495, opened 2018-05-28
Fix db:seed applications creation

calj:


db/seeds.rb
```diff
@@ -66,7 +66,7 @@
 logger.info 'Seeding applications'
 seeds['applications'].each do |seed|
   logger.info '---'
-  if Doorkeeper::Application.find_by(uid: seed['name']).present?
```
maksim-litvinov: It should be
```
Doorkeeper::Application.find_by(name: seed['name']).present?
```
The point is we check the name to prevent duplication of applications. `uid` field is dynamic anyway, we can't freeze it

PR #537, opened 2018-06-14
Add account uid to system.notification.account

maksim-litvinov:


mod:
@maksim-litvinov shouldn't we use uid instead of account_uid ?
also uid means user-id

app/mailers/account_mailer.rb
```diff
@@ -8,12 +8,14 @@ def confirmation_instructions(record, token, opts = {})
   end
 
   def reset_password_instructions(record, token, opts = {})
-    EventAPI.notify('system.notification.account', reset_password_token: token)
+    EventAPI.notify('system.notification.account',
```
geoffw8: I think the naming keys here are the wrong way round, it should actually be `system.account.notification` right?

Also - how comes this isn't actually `system.account.forgotten_password`?

PR #542, opened 2018-06-18
Compute code coverage report

calj:


Gemfile
```diff
@@ -58,7 +58,10 @@ group :test do
   gem 'shoulda-matchers',    '~> 3.1.2'
   gem 'rails-controller-testing'
   gem 'capybara-screenshot'
-  gem 'simplecov', require: false
+  gem 'simplecov'
```
maksim-litvinov: we don't need these gems in production, add `require 'false'` for all

PR #548, opened 2018-06-21
Feature/login history

tamadamas:


db/migrate/20180612084542_create_device_activity.rb
```diff
@@ -0,0 +1,17 @@
+class CreateDeviceActivity < ActiveRecord::Migration[5.1]
+  def change
+    create_table :device_activity do |t|
+      t.references :account, foreign_key: true, index: true, null: false
+      t.string :device_uid
```
mod: rename table to devices
device.uuid
device.expire_at
add otp columns
rename status in result

PR #601, opened 2018-09-18
Fix message error

mnaichuk:


rxx:
@mnaichuk It should be also for 1-8-stable branch

PR #603, opened 2018-09-18
Add Authz endpoint to auth Ambasador requests

rxx:


app/controllers/authz_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    'accounts/sign_out',
+    'accounts/sign_up',
+    'health/alive',
+    'health/ready'
+  ]
+
+  def verify
+    return head(:ok) if public_route?
+
+    response.set_header('Authorization', "Bearer #{create_token}")
```
mod: @rxx can you verify session is active

app/controllers/authz_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    head :ok
+  end
+
+  private
+
+  def create_token
```
mod: Could you use a non Doorkeeper method since we are going to delete this?
I suggest using gem peatio

config/application.rb
```diff
@@ -23,6 +23,16 @@ class Application < Rails::Application
     config.generators.test_framework :rspec
     config.eager_load_paths += %W[#{config.root}/lib]
```
mod: I don't think this relate to this patch ?
why do we need CORS in authz component ?

PR #604, opened 2018-09-18
Add ability to override url root path

shal:


tamadamas (approved):

ec:
Added into apigate story.
Please follow branch feature/apigate

PR #607, opened 2018-09-19
#PPBB-45: Allow to configure multiple CORS at barong

rxx:


app/api/user_api/v1/cors/middleware.rb
```diff
@@ -16,27 +16,32 @@ def call(headers)
...
-          {
-            'Access-Control-Allow-Origin'      => ENV.fetch('API_CORS_ORIGINS', '*'),
+        def headers(origin)
+          result = {
+            'Access-Control-Allow-Origin'      => '*',
             'Access-Control-Allow-Methods'     => 'GET, POST, PUT, PATCH, DELETE',
             'Access-Control-Allow-Headers'     => 'Origin, X-Requested-With, Content-Type, Accept, Authorization',
             'Access-Control-Allow-Credentials' => ENV['API_CORS_ALLOW_CREDENTIALS'].present?.to_s
           }
+          return result if ENV['API_CORS_ORIGINS'].blank? || ENV['API_CORS_ORIGINS'] == '*'
```
kovetskiy: We should not allow users to omit whitelists, whitelist must be mandatory because it's very insecure to not restrict Access-Control-Allow-Origin, especially for such services as barong.

kovetskiy: Also, don't allow users to specify "*", it's insecure by its nature.

PR #614, opened 2018-09-24
Add Authz endpoint to auth Ambasador requests

tamadamas:
Use jwt instead of doorkeeper


ec:
Feature moved to another branch. Follow feature/apigate branch

PR #627, opened 2018-10-12
Allow to configure root domain in development

tamadamas:
Workbench works in development mode


tamadamas:
moved to this pr https://github.com/rubykube/barong/pull/629

PR #632, opened 2018-10-16
Add RabbitMQ middleware for event api

tamadamas:


ec:
We should initialize ```EVENT_API_JWT_PRIVATE_KEY```, as it use all over the PR.

ec:
duplicated #640

PR #641, opened 2018-10-24
Master fixes

ec:


ec:
Source of the problem was in the Faker.image method.
Fixed from Faker side

PR #649, opened 2018-11-19
Fix account unlocking (account should be unlocked after 1 hour by default)

m-an:
+ Account unlocks after 1 hour by default
+ Add ability to configure unlock time with `"UNLOCK_IN"` env variable.


calj (changes requested):
Is there any change in DB schema?
Please clean this PR

calj (approved):

PR #655, opened 2018-11-25
Feature/seeder

calj:


lib/barong/seed.rb
```diff
@@ -0,0 +1,108 @@
...
+        ERB.new(
+          File.read(
+            ENV.fetch('SEEDS_FILE', Rails.root.join('config', 'seeds.yml'))
+          )
+        ).result
+      )
+    end
+
+    def show_seeded_users
+      puts 'Seeded users:'
```
mod: I suggest not printing from lib but you can return a string

lib/barong/seed.rb
```diff
@@ -0,0 +1,108 @@
...
+          logger.info "Created user for '#{user.email}'"
+
+          # Set correct level with labels
+          levels = levels = Level.where(id: 1..user.level)
+          raise ConfigError.new("No enough levels found in database to grant the user to level #{user.level}") if levels.count < user.level
+          levels.find_each do |level|
+            user.add_level_label(level.key, level.value)
+          end
+
+          # Confirm the email
```
mod: I think label was added before L68
and this code doesn't make sense to me
also we don't have a field updated_at

lib/barong/seed.rb
```diff
@@ -0,0 +1,108 @@
...
+          if user.update(updated_at: Time.current)
+            user.add_level_label('email')
+            logger.info("Confirmed email for '#{user.email}'")
+          end
+
+          # Create a Profile using defaults where values are not set in seeds.yml
+          if seed['phone']
```
mod: I suggest to remove this feature, never used it never will.
better we start using DB dumps for QA

lib/barong/seed.rb
```diff
@@ -0,0 +1,108 @@
...
+
+            if phone.save && phone.update(validated_at: Time.current)
+              logger.info "Created phone for '#{user.email}'"
+            else
+              logger.error "Can't create phone for '#{user.email}': #{phone.errors.full_messages.join('; ')}"
+            end
+          end
+
+          # Create a Profile using defaults where values are not set in seeds.yml
+          if seed['profile']
```
mod: remove profile, anyway it's a non-sense to seed a fake profile

mod (changes requested):

PR #665, opened 2018-12-02
GeeTest.com captcha integration on Barong v2

bsuprun:


ec:
related to PR

https://github.com/bodiasuprun/geetest-ruby

ec:
So, the plan now is more clear:

We should have a configuration `captcha_policy` with 3 parameters
> disabled
> geetest_captcha
> re_captcha

Depending on that we can improve our verify and API parameters logic.
See the comments on commit for detailed plan

app/api/v2/identity/sessions.rb
```diff
@@ -20,6 +20,11 @@ class Sessions < Grape::API
                                         desc: 'Response from Recaptcha widget'
           optional :otp_code, type: String,
                               desc: 'Code from Google Authenticator'
+          optional :geetest_response, type: Hash do
```
ec: Here we can leave only optional `:captcha_response, type: any`
No need to specify exact type and requires, as we dont need the captcha type at this point

app/api/v2/identity/sessions.rb
```diff
@@ -28,6 +33,8 @@ class Sessions < Grape::API
           verify_captcha!(user: user,
                           response: params['recaptcha_response']) if declared_params[:recaptcha_response]
 
+          verify_geetest!(response: params['geetest_response']) if declared_params[:geetest_response]
```
ec: here should be helper method `verify_captcha`
In the helpers file (utils.rb) you can check the configuration.
> skip if disabled
>  if its recaptcha - call recaptcha_verifier
>  if its geetest -  check if all the params are present (challenge, validate, seccode) and call geetest verifier

app/api/v2/identity/users.rb
```diff
@@ -17,17 +17,31 @@ class Users < Grape::API
         params do
           requires :email, type: String, desc: 'User Email', allow_blank: false
           requires :password, type: String, desc: 'User Password', allow_blank: false
-          requires :recaptcha_response, type: String, desc: 'Response from Recaptcha widget'
+          optional :recaptcha_response, type: String, desc: 'Response from Recaptcha widget'
```
ec: here is the same

app/api/v2/identity/users.rb
```diff
@@ -17,17 +17,31 @@ class Users < Grape::API
...
+                          response: params['recaptcha_response']) if declared_params[:recaptcha_response]
+
+          verify_geetest!(response: params['geetest_response']) if declared_params[:geetest_response]
 
           error!(user.errors.full_messages, 422) unless user.save
         end
 
+        desc 'Register Geetest captcha'
+        get '/register_geetest' do
+          GeetestVerifier.new.register()
```
ec: Should be in helpers and executed automatically if its geetest

app/services/geetest_verifier.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require 'digest'
+require 'net/http'
+
+class GeetestVerifier
```
ec: We should look for the opportunity to merge both types of captcha in one service file

app/services/geetest_verifier.rb
```diff
@@ -0,0 +1,42 @@
...
+  def initialize
+    @api = 'http://api.geetest.com'
+    @validate_path = '/validate.php'
+    @register_path = '/register.php'
+    @geetest_id = ENV['GEETEST_ID']
+    @geetest_key = ENV['GEETEST_KEY']
```
ec: This secrets should go to barong secret storage. Look over the app to see an example

ec (changes requested):

bsuprun:
TODO: fix tests

ec (approved):

ec:
Merge branch into master commit should be inside the PR

PR #688, opened 2018-12-14
Add change password endpoint

m-an:


app/api/v2/identity/users.rb
```diff
@@ -151,6 +151,30 @@ class Users < Grape::API
...
+                             allow_blank: false
+            requires :old_password, type: String,
+                                    desc: 'Old password',
+                                    allow_blank: false
+            requires :new_password, type: String,
+                                    desc: 'New password',
+                                    allow_blank: false
+          end
+
+          put do
```
ec: We will need an activity record on the action
Take a look on reset password, for example

app/api/v2/identity/users.rb
```diff
@@ -151,6 +151,30 @@ class Users < Grape::API
             EventAPI.notify('system.user.password.reset', current_user.as_json_for_event_api)
             status 201
           end
+
+          desc 'Change password' do
```
ec: Lets move this endpoint to resource module, as it should be a protected api

app/api/v2/identity/users.rb
```diff
@@ -151,6 +151,30 @@ class Users < Grape::API
...
+            success { code: 201, message: 'Changed' },
+            failure [
+              [ 401, 'Unauthorized' ],
+              [ 422, 'Unprocessable Entity' ],
+              [ 404, 'Not found' ]
+            ]
+          end
+
+          params do
+            requires :email, type: String,
```
ec: As its a protected API - we dont need an email in params, as this information is clear from JWT and current_user
Take a look at utils file in resource api module

ec (changes requested):

ec (approved):

PR #706, opened 2019-01-03
Add users entity on login and /me

ec:


app/api/v2/base.rb
```diff
@@ -33,7 +33,9 @@ class Base < Grape::API
       Entities::Label,
       Entities::APIKey,
       Entities::Profile,
-      Entities::UserWithProfile
+      Entities::User,
+      Entities::UserWithProfile,
```
mod, line 37: are you sure to have defined       Entities::UserWithFullInfo ? @ec

ec: @mod  Yes, I am. It's in entities since adding admin module.

We really need such kind of entity for admin panel, as on old UI on one page (user show) we had everything about user. It saves a lot of requests for front devs.

PR #709, opened 2019-01-08
Add referral_id to users table and as an optional param on signup

ec:


app/api/v2/identity/users.rb
```diff
@@ -16,12 +16,15 @@ class Users < Grape::API
...
           requires :email, type: String, desc: 'User Email', allow_blank: false
           requires :password, type: String, desc: 'User Password', allow_blank: false
+          optional :referral_id, type: String, desc: 'Referral uid'
           optional :captcha_response, types: [String, Hash],
                                       desc: 'Response from captcha widget'
         end
         post do
           declared_params = declared(params, include_missing: false)
           user_params = declared_params.slice('email', 'password')
+          user_params[:referral_id] = params[:referral_id] if params[:referral_id]
```
mod: we need to regex validate that it's a UID, I think they have a fixed size of 10 and stars with ID
ID[0-9]{8}

app/models/user.rb
```diff
@@ -80,11 +80,20 @@ def as_json_for_event_api
   end
 
   def as_payload
-    as_json(only: %i[uid email role level state])
+    as_json(only: %i[uid email referral_id role level state])
   end
 
   private
 
+  def validate_and_assign_ref_id
```
mod: I disagree with the approach
Referral ID should be simply like email is set and saved, validated.
but there is no relationship between generating a UID and setting a referral ID.
we just need to validate it's correct format, and it exist in db.

mod (changes requested):

app/api/v2/identity/users.rb
```diff
@@ -5,6 +5,20 @@
...
+        def validate_refid?
+          return false if params[:refid].nil?
+
+          error!('Invalid referral uid format', 422) unless /\AID\w{10}$/.match?(params[:refid])
+          error!("Referral doesn't exist", 422) if User.find_by_uid(params[:refid]).nil?
+          true
+        end
+
+        def referral_id
+          User.find_by_uid(params[:refid]).id
```
mod: what happens if it doesn't find UID ?

app/api/v2/identity/users.rb
```diff
@@ -5,6 +5,20 @@
 module API::V2
   module Identity
     class Users < Grape::API
+      helpers do
+        def validate_refid?
+          return false if params[:refid].nil?
```
mod: maybe validate_refid!()

app/api/v2/identity/users.rb
mod: try to make functions bigger than 1 lines

app/api/v2/identity/users.rb
```diff
@@ -16,12 +30,16 @@ class Users < Grape::API
...
           requires :password, type: String, desc: 'User Password', allow_blank: false
+          optional :refid, type: String, desc: 'Referral uid'
           optional :captcha_response, types: [String, Hash],
                                       desc: 'Response from captcha widget'
         end
         post do
           declared_params = declared(params, include_missing: false)
           user_params = declared_params.slice('email', 'password')
+
+          user_params[:referral_id] = referral_id if validate_refid?
```
mod: My suggestion:

`user_params[:referral_id] = parse_refid! unless params[:refid].nil?`

combine into 1 function, improve clarity

db/migrate/20190108115333_add_referral_id_to_users_table.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Add new field due to referral - affiliate relation
+class AddReferralIdToUsersTable < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :referral_id, :integer, after: :password_digest
```
mod: user_id is bitint not integer,
also you can place it at the end of the table before timestamps

mod (changes requested):

app/api/v2/identity/users.rb
ec:
```
if User.find_by_uid(params[:refid]).nil?
error!("Referral doesn't exist", 422)
```

Squashed 2 funcs into one

mod (approved):

db/migrate/20190108115333_add_referral_id_to_users_table.rb
hardywu: would be better with  `add_index :users, :referral_id`

PR #716, opened 2019-01-09
Add an ability to specify CORS(multiple)

ec:


mod (changes requested):
I prefer using https://github.com/ruby-grape/grape#cors
Rack::Cors middleware as it's natively supported by grape
I should be smart behavior with OPTIONS requests.

PR #721, opened 2019-01-11
add user.referees/user.referrer and referral in api entity

hardywu:


app/models/user.rb
```diff
@@ -25,6 +25,8 @@ class User < ApplicationRecord
   scope :active, -> { where(state: 'active') }
 
   before_validation :assign_uid
+  belongs_to :referrer, class_name: "User", optional: true, foreign_key: 'referral_id'
```
ec: ruby prefer single quote syntax

app/models/user.rb
```diff
@@ -25,6 +25,8 @@ class User < ApplicationRecord
...
+  has_many :referees, class_name: "User", foreign_key: 'referral_id'
```
ec: the same

app/api/v2/entities/user_with_full_info.rb
```diff
@@ -13,6 +13,7 @@ class UserWithFullInfo < Grape::Entity
         expose :level, documentation: { type: 'Integer' }
         expose :otp, documentation: { type: 'Boolean' }
         expose :state, documentation: { type: 'String' }
+        expose :referral_id, documentation: { type: 'Integer' }
```
ec, line 16: changes in entity always cause error in related specs. fix the specs

ec (changes requested):

ec:
What is the purpose of returning referral_id on current stage?
What is the purpose of making a relations between referrals and referees? As barong doesnt have any in-app functionality for that, and doesn't have api to use it

hardywu:
@ec  The Frontend need to know whether the referral goes through successfully, particularly if there is a reward notification for a referral.

ec:
@hardywu 
But still, you dont add an api for relation logic. So this part of code is "dead" now, which shouldn't be here.

Then, about referral_id. Referral_id is an integer of user id.
ID comes from DB and we don't return it in API. (Only UID, which is identifier for frontend)

Still, with referral_id front still doesnt know anything about referrer

hardywu:
how about using a Boolean attribute `referred: true/false`

ec:
Well, i doesn't see any advantage of it. If frontend still dont know about WHO is referrer - what is the purpose of knowledge user is referred?

hardywu:
it's not about WHO is referrer, but letting front know the referral go through. A user would never know if his referral_id is correctly typed on registration, nor can he/she notify his friend who referred him.

ec:
Still, its not the functionality, that should be in the barong. At least, now.
I guess, we will come back to this question a bit later, when peatio or core frontend functionality will strongly require it

PR #731, opened 2019-01-17
Move session#logout to resource module

### LINKED ISSUE #733: Invalid session cookies returns on session logout

ec:
* Make session#logout protected api
* Stop send invalid cookie on logout

closes #733


mod:
I would highly prefer to bring auth in idenity/session logout here
also we don't need to necessarly protect this endpoint session must be valid anyway.

ec:
Freezed

PR #745, opened 2019-01-28
Added event API documentation with examples

### LINKED ISSUE #735: Event api doc v2

ec:
closes #735

### DIFF
--- a/docs/event_api.md
+++ b/docs/event_api.md
@@ -1,0 +1,273 @@
+# RabbitMQ Barong Event API
+
+## Overview of RabbitMQ details
+
+Barong submits all events into three exchanges depending on event category (read next).
+
+The exchange name consists of three parts: 
+
+  1) application name (typically `barong`)
+
+  2) fixed keyword `events`.
+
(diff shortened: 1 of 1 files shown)


app/api/v2/identity/users.rb
```diff
@@ -172,7 +172,7 @@ def parse_refid!
 
             activity_record(user: current_user.id, action: 'password reset', result: 'succeed', topic: 'password')
 
-            EventAPI.notify('system.user.password.reset', current_user.as_json_for_event_api)
+            EventAPI.notify('system.user.password.reseted', current_user.as_json_for_event_api)
```
mod: reset in the past is reset too
I don't mind the convention that was before using present everywhere

shal (approved):

mod (changes requested):

app/api/v2/identity/users.rb
mod: rename to present time

mod (changes requested):

PR #765, opened 2019-02-14
Simplify, make more readable user agent attribute

ec:


mod:
I would put the parser in client side, I think you are changing the level of information we keep for lower detail

ec:
frozen

PR #766, opened 2019-02-14
Search functionality for admin module

ec:


mod:
I think it's not the correct feature, we are looking for ability to search users we should make a search endpoint which returns a list of user matching certain params

ec:
reworked #769

PR #769, opened 2019-02-15
Search implementation (admin api)

ec:


app/api/v2/admin/utils.rb
```diff
@@ -0,0 +1,19 @@
...
+
+module API
+  module V2
+    module Admin
+      # helper module
+      module Utils
+        def search(field, value)
+          case field
+          when 'email'
+            User.where("users.#{field} LIKE ?", "#{value}%")
```
mod: 1 file for 1 line of code?

app/api/v2/admin/users.rb
```diff
@@ -21,6 +21,23 @@ class Users < Grape::API
...
+          failure: [
+            { code: 401, message: 'Invalid bearer token' }
+          ]
+          params do
+            requires :field,    type: String, desc: 'User model field.'
+            requires :value,    type: String, desc: 'First part of a value (search target)'
+            optional :page,     type: Integer, default: 1,   integer_gt_zero: true, desc: 'Page number (defaults to 1).'
+            optional :limit,    type: Integer, default: 100, range: 1..1000, desc: 'Number of withdraws per page (defaults to 100, maximum is 1000).'
+          end
+          get '/search' do
```
mod, line 43: we need to be able to search in more fields from profile

mod (changes requested):

mod (changes requested):
Don't create a file for this search feature
Put it in the model.
Try to support search by field using =
Like state, role, email, uid

Support search LIKE `value%`
On most field of profile maybe address, country, city as well.

mod (approved):

PR #771, opened 2019-02-18
Add pagination on /resource/users/activity endpoint

ec:


mod:
are we using the same pagination as in peatio?
kaminari + api-paginate

ec:
> are we using the same pagination as in peatio?
> kaminari + api-paginate

yes, we do

mod (approved):

PR #774, opened 2019-02-19
Feature/validation errors

ec:


mod:
What are the changes in the gem why not copying it in lib/strong_password instead .

Maybe make a new repository ?

PR #781, opened 2019-02-25
Implement api keys scope logic

### LINKED ISSUE #779: Implement api keys scope logic

ec:
closes #779

### DIFF
--- a/app/api/v2/resource/api_keys.rb
+++ b/app/api/v2/resource/api_keys.rb
@@ -30,10 +30,10 @@ module API::V2
           optional :kid,
                    type: String,
                    allow_blank: false
-          optional :scope,
-                   type: String,
+          optional :json,
+                   type: JSON,
                    allow_blank: false,
-                   desc: 'comma separated scopes'
+                   desc: 'json string with key: value pair'
           requires :totp_code,
                    type: String,
                    message: 'resource.api_key.missing_totp',
@@ -43,7 +43,6 @@ module API::V2
         post do
           declared_params = declared(unified_params, include_missing: false)
                             .except(:totp_code)
-                            .merge(scope: params[:scope]&.split(','))
           api_key = current_user.api_keys.create(declared_params)
           if api_key.errors.any?
             code_error!(api_key.errors.details, 422)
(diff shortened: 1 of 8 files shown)


ec:
@mod here is implementation of api keys scope logic:

### Rules

Rules fetches from `/config/authz_rules.yml` and typically should look:
```ruby
api key rules:
  scope_name: 
    - api/v2/allowed-path-for-this-scope
    - api/v2/allowed-path-for-this-scope
    ...
  scope_name:
    - api/v2/allowed-path-for-this-scope
    ...
  scope_name:
    - api/v2/allowed-path-for-this-scope
    ...
```

Although i've put the default scopes `account, trade, withdraw` inside, it can be changed by mounting custom yml in image

P.S. remember, that no need to specify the full path for every API endpoint, enough to specify module name or namespace ( as check is `starts_with?`)

### Read and Write access
Now we use `serialize :scope, JSON` in model (previously it was array)

So, now each scope can have `read` or `write` value, 
(ex. `scope: {"withdraw"=>"read"`)

Depending on the value, AuthZ will allow all the methods (POST, PUT, DELETE etc) for scopes with `write` access, and only GET requests for scope with `read` access

P.S. multiple scopes allowed for one api key.
(ex. `scope: {"withdraw"=>"read", "trade"=>"write"}`)

### Errors

As it shown in the spec in the patch, if api key tries to access the path, that is not allowed for current scope AuthZ will return error:
status: `401`
body: `{ errors: [authz.api_key_not_enough_rights] }`

The same error will be returned if scope and path fits each other, but api key trying to do a POST with READ scope access

app/api/v2/resource/api_keys.rb
```diff
@@ -33,7 +33,7 @@ class APIKeys < Grape::API
           optional :scope,
```
ysv: @ec you can specify this parameter type as JSON
https://github.com/ruby-grape/grape#first-class-json-types 
This should give an ability to validate scopes on fly

config/authz_rules.yml
```diff
@@ -2,10 +2,24 @@
 # pass for whitelisted (public) routes
 # block for blacklisted routes
 #
-rules:
+general rules:
```
ysv: I would prefer underscore version `general_rules`

ec:
Frozen

PR #791, opened 2019-03-01
Improve language support in event API, add domain info inside event

### LINKED ISSUE #792: add domain info inside events

ec:
closes #792


mod:
this is too unsecure just make a variable BARONG_DOMAIN

ec:
It's not the barong domain, but frontend domain @mod 
As postmaster will send an email to front.

And as we are talking about adding the variable, then better to set it in postmaster.

PR #794, opened 2019-03-02
API user activity sort data by created_at

huuhait:


app/api/v2/resource/users.rb
```diff
@@ -34,7 +34,7 @@ def validate_topic!(topic)
         end
         get '/activity/:topic' do
           validate_topic!(params[:topic])
-          data = current_user.activities
+          data = current_user.activities.sort_by {|t| -t.created_at.to_i }
```
ec: Better to use `.order('created_at DESC')`, so it will look like

```ruby
data = current_user.activities.order('created_at DESC')
```

ec:
Please reply if you are going to fix this fast, or i will repush it by my own

ec (changes requested):

app/api/v2/resource/users.rb
huuhait: updated

app/api/v2/resource/users.rb
```diff
@@ -34,7 +34,7 @@ def validate_topic!(topic)
         end
         get '/activity/:topic' do
           validate_topic!(params[:topic])
-          data = current_user.activities
+          data = current_user.activities.order('created_at DESC')
```
ec, line 37: Can you also please add a spec on it?
Something simple, just `it returns activities from new to old`

ec (changes requested):

app/api/v2/resource/users.rb
huuhait: i'm poor rspec :|

ec:
duplicates #804

Issue #799, opened 2019-03-07
Admin api list users with pending documents


ec:

ec:
already in master

PR #800, opened 2019-03-07
Add ability to search users by labels

mkalenska:


ec:
closes #799

ec (approved):

app/api/v2/admin/users.rb
```diff
@@ -225,6 +225,24 @@ def search(field, value)
...
+              { code: 401, message: 'Invalid bearer token' }
+            ]
+            params do
+              requires :key,      type: String, desc: 'Label key'
+              requires :value,    type: String, desc: 'Label value'
+              optional :page,     type: Integer, default: 1,   integer_gt_zero: true, desc: 'Page number (defaults to 1).'
+              optional :limit,    type: Integer, default: 100, range: 1..1000, desc: 'Number of users per page (defaults to 100, maximum is 1000).'
+            end
+            get '/search' do
+              users = User.joins(:labels).where(labels: { scope: 'private', key: params[:key], value: params[:value] })
```
mod: I think we need to remove private scope from the search, label can be public
I think we can improve this api

mod (changes requested):

app/api/v2/admin/users.rb
ec: then, should we add a new optional parameter to specify?
Cause mixing private and public is not so good idea

mod: ok, sounds good but maybe /search is not the right name or maybe we could do more search possibilities ?

mod (changes requested):

PR #803, opened 2019-03-09
Expose Referral UID in API

### LINKED ISSUE #802: Expose Referral ID in API
Expose Referral ID in GET methods for User and Management API

martianatwork:
This should close #802 
Using UID instead of ID because the ID will be different in different systems i.e peatio.


ec:
Hello @martianatwork . Thank you for contributing, but topic is not so clear as you think. Main functionality should go to management API, not for user one.
Please wait a bit, I will post an update and patch in issue comments.

I'll close PR. For sure, we will not implement the story like this.

PR #805, opened 2019-03-11
Move event api jwt private key to Barong::App.config store

ec:


mod:
maybe we use credentials.yml ?

mod:
I don't understand this path, we need to make sure we use this key only in development,
maybe it's even not a good idea to commit a key at all.

I was thinking credentials in dev and test isn't so bad in this case

ec:
@mod so, i consider we put the default for event_api_jwt_private_key inside credentials, but use in lib `Barong::App.config`, as it will try to 
1) fetch value from env (which is ok for test and dev mode)
2) If no env provided - check inside credentials (which is the best solution for dev mode)
3) Throw error if not configured by env on production mode.

Issue #813, opened 2019-03-18
Feature Request: Configurable roles and permissions


ec:

ec:
Available in 2.2

PR #814, opened 2019-03-18
Add Permission model, permissions seed

ec:


ec:
closes #813

ec:
duplicates #819

PR #815, opened 2019-03-21
Add opportunity to get list of users by management API

ymasiuk:


spec/api/v2/management/users_spec.rb
```diff
@@ -100,6 +100,51 @@
...
+      let(:data) do
+        {
+          scope: :read_users
+        }
+      end
+      let(:signers) { %i[alex jeff] }
+
+      let(:do_request) do
+        post_json '/api/v2/management/users/list',
+                  multisig_jwt_management_api_v1({ data: data }, *signers)
```
ec: It's actually should be `multisig_jwt_management_api_v2`

ymasiuk: @ec done

ec (approved):

ec (approved):

PR #819, opened 2019-04-15
Permissions & roles, seed feature and authz logic

ec:


ec:
Probably permissions should be divided like this

DeepinScreenshot_select-area_20190418124121

mod (approved):

mod (approved):

PR #823, opened 2019-04-22
Make :doc_expire an optional field

ec:


calj (changes requested):
Please make it configurable, so we can allow or not docs to have no expiration date.

config/initializers/document.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+Barong::App.define do |config|
+  config.set(:optional_docs_expire, false)
```
calj: let's rename it required_docs_expire

app/api/v2/resource/documents.rb
```diff
@@ -38,10 +34,18 @@ class Documents < Grape::API
...
+          optional :doc_expire,
+                   type: { value: Date, message: "resource.documents.expire_not_a_date" },
+                   allow_blank: false,
+                   desc: 'Document expiration date'
           optional :metadata, type: Hash, desc: 'Any key:value pairs'
         end
 
         post do
+          if Barong::App.config.required_docs_expire
+            error!({ errors: ['resource.documents.invalid_format'] }, 422) unless /\A\d{4}\-\d{2}\-\d{2}\z/.match?(params[:doc_expire].to_s)
```
calj, line 46: we should probably check that the date is in future

PR #824, opened 2019-05-06
Add GET /labels/list to return key, value, scope of all existing labels

ec:


ec:
`Label.where(scope: 'private').select(:key, :value).group(:key, :value).as_json.map(&:compact)`

```
SELECT `labels`.`key`, `labels`.`value` FROM `labels` WHERE `labels`.`scope` = 'private' GROUP BY `labels`.`key`, `labels`.`value`
```


response example: 
```
[{"key"=>"document", "value"=>"verified"}, {"key"=>"email", "value"=>"verified"}, {"key"=>"phone", "value"=>"verified"}]
```

app/api/v2/admin/users.rb
```diff
@@ -101,6 +101,21 @@ def search(field, value)
...
           namespace :labels do
+            desc 'Returns existing labels keys and values',
+            security: [{ "BearerToken": [] }],
+            failure: [
+              { code: 401, message: 'Invalid bearer token' }
+            ]
+            params do
+            end
+            get '/list' do
+              labels = Label.where(scope: 'private').pluck(:key, :value)
```
ec: Label.where(scope: 'private').group(:key, :value).size

PR #825, opened 2019-05-06
Add user self full info endpoint in resource module

ec:


ec:
will be split in several endpoints

PR #826, opened 2019-05-09
Added API call to get users with pending documents

gfedorenko:


app/api/v2/admin/users.rb
```diff
@@ -119,6 +119,22 @@ def search(field, value)
...
+                 security: [{ "BearerToken": [] }],
+                 failure: [
+                     { code: 401, message: 'Invalid bearer token' }
+                 ]
+            params do
+              optional :page,     type: Integer, default: 1,   integer_gt_zero: true, desc: 'Page number (defaults to 1).'
+              optional :limit,    type: Integer, default: 100, range: 1..1000, desc: 'Number of users per page (defaults to 100, maximum is 1000).'
+            end
+            get '/pending' do
+              users = User.joins(:labels).where(labels: { key: 'document', value: 'pending' })
```
ec: Can you post resulting SQL query pls

gfedorenko: `SELECT  `users`.* FROM `users` INNER JOIN `labels` ON `labels`.`user_id` = `users`.`id` WHERE `labels`.`key` = 'document' AND `labels`.`value` = 'pending' LIMIT 100 OFFSET 0`

ec: We need in this list only system ones, as labels has a scope, and public one can be added by users by themselves 

So, add a `scope: private` in where clause
And, maybe, a spec testing it returns only private ones 
And a spec testing it returns only private one

ec (changes requested):

ec (approved):

ec:
@gfedorenko lets be more explicit and order users by labels `updated_at` field

PR #828, opened 2019-05-15
Added admin endpoints to get activities

gfedorenko:


ec (approved):

pkucherenk0:
@gfedorenko Pls add new calls to documentation.

pkucherenk0:
Why do we filter action instead of topic as we have in public api?
https://github.com/rubykube/barong/blob/master/docs/rest_api.md#resourceusersactivitytopic

Ability to get all user activities will require to return which activity belong to which user.

ec:
@gfedorenko https://ci.microkube.com/rubykube/barong/457/1/6

ec (approved):

pkucherenk0:
I have problems with starting barong docker image with latest commit.
Selection_339
____________________________
Was issue on my side.

pkucherenk0:
1. The sorting of activities should be from from newest to oldest. (Admin would like to check resent activity on the first page over first login after registration)
2. `created_from` should be  `=` and `>` from inputted date and `created_to` should be `=` and `<` from inputted date. (At the moment it is `>` and `<` the imputed date vs time will be excluded)

ec:
@pkucherenk0 approved, will be updated soon

PR #829, opened 2019-05-15
Disable APIKeys when user state or otp changes

dnfd:


ec (approved):

app/api/v2/admin/users.rb
```diff
@@ -88,14 +88,20 @@ def search(field, value)
...
-            if update_param_key == 'role' && update_param_value == true
+            if update_param_key == 'otp' && update_param_value == true
               error!({ errors: ['admin.user.enable_2fa'] }, 422)
             end
 
             if update_param_value == target_user[update_param_key]
               error!({ errors: ["admin.user.#{update_param_key}_no_change"] }, 422)
             end
 
+            if update_param_key == 'state' || update_param_key == 'otp'
```
calj: let's move this logic to User model after_update

app/models/user.rb
```diff
@@ -25,6 +25,15 @@ class User < ApplicationRecord
   scope :active, -> { where(state: 'active') }
 
   before_validation :assign_uid
+  after_update :disable_api_keys
+
+  def disable_api_keys
```
calj, line 29: It looks like we actually don't check the otp or user state

PR #830, opened 2019-05-15
Add ability to get list of users with profile

gfedorenko:


ec (approved):

spec/api/v2/admin/users_spec.rb
```diff
@@ -55,6 +56,26 @@ def validate_fields(user)
...
+      end
+      let(:do_extended_info_request) { get '/api/v2/admin/users', headers: auth_header, params: extended_params }
+
+      it 'returns list of users with full info' do
+        do_extended_info_request
+        users = JSON.parse(response.body)
+        expect(User.count).to eq users.count
+
+        expect(users.first['labels']).to eq []
+        expect(users.first['activities']).to eq []
```
calj: We need remove activities from this call.

gfedorenko: @calj  So we should remove them in entity?

spec/api/v2/admin/users_spec.rb
```diff
@@ -41,6 +41,7 @@
       def validate_fields(user)
         user.attributes.slice('email', 'role', 'level', 'otp', 'state', 'uid')
       end
+
       it 'returns list of users' do
         do_request
         users = JSON.parse(response.body)
```
calj: Assert here that extended informations are missing.

spec/api/v2/admin/users_spec.rb
ec: @gfedorenko no, we need to add additional entity.
Activities inside this entity used in several endpoints, so we cant remove them

pkucherenk0:
@gfedorenko Pls add new calls to documentation.
Functional side looks fine by me.

ec:
@pkucherenk0 documentation will be updated automatically (if kite will not die again)

ec (approved):

mod (approved):

PR #852, opened 2019-06-23
Split and refactor auth spec, add sync auditor for test env

ec:


lib/barong/activity_logger.rb
```diff
@@ -21,6 +21,12 @@ def self.write(options = {})
       end
     end
 
+    def self.sync_write(options = {})
+      Activity.create(format_params(options))
+    rescue => exception
```
calj: I would not rescue any exception here, especially this will be used in tests, so better we brake instead of logging somewhere the exception

PR #856, opened 2019-06-24
Add docs in bump and update redeploy step

dnfd:


ec (approved):
awesome

ec:
@calj here is result (tested on branches)
https://github.com/rubykube/barong/commit/bd656597e737d7ca235f44302a5b8e9580c1c008

PR #860, opened 2019-06-27
Move from carrierwave to Active Storage support

ec:


ec:
frozen.
Will reopen as soon as any action will be done

PR #863, opened 2019-07-02
Update hyperlink

liutenko:


ec:
I will include it in another patch. Just for not pushing so minor changes

PR #864, opened 2019-07-02
Add filter/users in labels management api module

ec:


app/api/v2/management/labels.rb
```diff
@@ -1,83 +1,132 @@
...
+                   type: { value: Integer, message: 'Non integer limit' },
+                   values: { value: 1..1000, message: 'Invalid limit' },
+                   default: 100,
+                   desc: 'Number of users per page (defaults to 100, maximum is 1000).'
+        end
+        post '/filter/users' do
+          entity = params[:extended] ? API::V2::Entities::UserWithProfile : API::V2::Entities::User
+          users = API::V2::Queries::UserWithLabelFilter.new(User.all).call(permitted_search_params(params))
+
+          users.tap { |q| present paginate(q), with: entity }
```
denisfd: ```present paginate(users), with: entity```

app/api/v2/management/labels.rb
```diff
@@ -1,83 +1,132 @@
...
+          success API::V2::Entities::Label
+        end
+        params do
+          requires :user_uid, type: String, allow_blank: false, desc: 'User uid'
+          requires :key, type: String, allow_blank: false, desc: 'Label key.'
+        end
+        post '/delete' do
+          label = user.labels.find_by!(key: params[:key], scope: 'private')
+          label.destroy
+          status 204
```
denisfd: empty line

app/api/v2/queries/user_with_label_filter.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    # initialize query to get User.all
+    def initialize(initial_scope)
+      @initial_scope = initial_scope.left_outer_joins(:labels)
+    end
+
+    # returns query with with all applied filters
+    def call(params)
+      scoped = filter_by_date(initial_scope, params[:range], params[:from], params[:to])
+      scoped = filter_by_key(scoped, params[:key])
```
denisfd: https://gist.github.com/denisfd/01cb2da9f04a9119da1c721df4fa8de8

dnfd, line 14: a bit shorter https://gist.github.com/denisfd/01cb2da9f04a9119da1c721df4fa8de8

app/api/v2/management/labels.rb
dnfd, line 127: add empty line

app/api/v2/management/labels.rb
dnfd: ```present paginate(users), with: entity```

app/api/v2/queries/user_with_label_filter.rb
ec: @denisfd for sure, it is. But I prefer to keep this per function filtering.
As soon as we still build ONE query it is more about readability.

Your suggestion is OK, but it not what we likely want to have in every query filter class. If you check others (that already in master) you will see we can have up to 10 filter functions. And in this case your approach starts to work as "magic", which is not what we want to have in open source project

dnfd (approved):

mod (approved):

PR #875, opened 2019-07-23
Fix JWT payload on document create event

vpetrusenko:


ec (changes requested):
This should probably come with test, that prove a change

ec:
Outdated

Issue #876, opened 2019-07-25
Make phone code message configurable


ec:
https://github.com/rubykube/barong/blob/d5c9d4fbd79cb11ff0a4bdd5c3910ad80a0fe697/app/models/phone.rb#L45

ec:
closed via #877

PR #877, opened 2019-07-25
Add ability to configure phone verification sms content

ec:
* Add SMS_REARWARD and SMS_AFTERWARD envs & defaults for them
* Related unit tests


ec:
closes #876

mod:
I would prefer you make a variable

sms_template = "Your verification code is {{code}} please, merci thank you"
in ruby you substitute the s/{{code}}/235/g

mod:
I think there is a deprecated dead value to change barong it was an ENV

PR #882, opened 2019-07-29
Add ability to get referral uid && Expose referral_uid in user, user_with_profile, user_with_full_info entities

chumaknadya:


app/models/user.rb
```diff
@@ -44,6 +45,17 @@ def role_exists
     errors.add(:role, 'doesnt exist')
   end
 
+  def referral_exists
+    errors.add(:referral_id, 'doesnt exist') if referral_id.present? && User.find_by(id: referral_id).blank?
+  end
+
+  def referral_uid
+    user = User.find_by(id: referral_id)
```
ec, line 54: can be compacted into one-line-function

app/models/user.rb
```diff
@@ -44,6 +45,17 @@ def role_exists
     errors.add(:role, 'doesnt exist')
   end
 
+  def referral_exists
+    errors.add(:referral_id, 'doesnt exist') if referral_id.present? && User.find_by(id: referral_id).blank?
```
ec: we use dictionary errors, so `doesnt_exist` or just `exist` - it will point FE with translations

app/api/v2/entities/user_with_full_info.rb
```diff
@@ -17,6 +17,9 @@ class UserWithFullInfo < Grape::Entity
         expose :labels, using: Entities::Label
         expose :phones, using: Entities::Phone
         expose :documents, using: Entities::Document
+        expose :referral_uid, documentation: { type: 'String' } do |user|
```
ec: better to have `desc:` here, as it's a bit unclear place

app/models/user.rb
```diff
@@ -44,6 +45,17 @@ def role_exists
     errors.add(:role, 'doesnt exist')
   end
 
+  def referral_exists
```
ec, line 49: to have comment on top of function here will be great

ec (changes requested):

ec (approved):

mod (approved):

PR #883, opened 2019-07-31
Add configuration for session opening: pending_login_allowed(boolean)

ec:


mod (changes requested):
We can do a mapping level

Like pending level = 1,2,3

Where we let user define his review process

ec:
Frozen until finding better solution

PR #891, opened 2019-08-13
Add Restrictions

dnfd:


app/api/v2/admin/restrictions.rb
```diff
@@ -0,0 +1,92 @@
...
+    module Admin
+      # Admin functionality over restrictions table
+      class Restrictions < Grape::API
+        resource :restrictions do
+          desc 'Returns array of restrictions as a paginated collection',
+          security: [{ 'BearerToken': [] }],
+          failure: [
+            { code: 401, message: 'Invalid bearer token' }
+          ]
+          params do
```
ec, line 14: we will need to have filtering here
At least on scope and timestamp maybe

app/api/v2/admin/restrictions.rb
```diff
@@ -0,0 +1,92 @@
...
+            optional :page,
+                     type: { value: Integer, message: 'admin.user.non_integer_page' },
+                     values: { value: -> (p){ p.try(:positive?) }, message: 'admin.user.non_positive_page'},
+                     default: 1,
+                     desc: 'Page number (defaults to 1).'
+            optional :limit,
+                     type: { value: Integer, message: 'admin.user.non_integer_limit' },
+                     values: { value: 1..100, message: 'admin.user.invalid_limit' },
```
ec: Maximum limit should be `1000` instead of `100`

app/api/v2/admin/restrictions.rb
```diff
@@ -0,0 +1,92 @@
...
+      class Restrictions < Grape::API
+        resource :restrictions do
+          desc 'Returns array of restrictions as a paginated collection',
+          security: [{ 'BearerToken': [] }],
+          failure: [
+            { code: 401, message: 'Invalid bearer token' }
+          ]
+          params do
+            optional :page,
+                     type: { value: Integer, message: 'admin.user.non_integer_page' },
```
ec: Errors represents modules, so we should use
`admin.restriction.non_integer_page`

Maybe its good idea to have something general for param validation error, but its a breaking change for UI translation

app/api/v2/admin/restrictions.rb
```diff
@@ -0,0 +1,92 @@
...
+                     values: { value: -> (p){ p.try(:positive?) }, message: 'admin.user.non_positive_page'},
```
ec: error `restriction`

app/api/v2/admin/restrictions.rb
```diff
@@ -0,0 +1,92 @@
...
+                     default: 1,
+                     desc: 'Page number (defaults to 1).'
+            optional :limit,
+                     type: { value: Integer, message: 'admin.user.non_integer_limit' },
```
ec: error message

app/api/v2/admin/restrictions.rb
ec: error message

app/api/v2/admin/restrictions.rb
```diff
@@ -0,0 +1,92 @@
...
+
+          desc 'Create restriction',
+          security: [{ 'BearerToken': [] }],
+          failure: [
+            { code: 401, message: 'Invalid bearer token' }
+          ]
+          params do
+            requires :scope,
+                     allow_blank: false,
+                     values: { value: -> { Restriction::SCOPES }, message: 'admin.restriction.invalid_scope'}
```
ec, line 52: And here you've changed the error to correct! :p 
ctrl-c ctrl-v get endpoint? XD

app/models/activity.rb
```diff
@@ -28,3 +28,25 @@ def readonly?
     !new_record?
   end
 end
+
+# == Schema Information
```
ec: Schema info in models - great tip, but its not the topic of PR
I will appreciate if you open another PR just with this stuff

app/models/api_key.rb
```diff
@@ -31,4 +31,22 @@ def hmac?
   def active?
     self.state == 'active'
   end
-end
\ No newline at end of file
+end
+
+# == Schema Information
```
ec: the same in all models

spec/factories/restriction.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :restriction do
```
ec, line 4: this looks suspicious
Always put a defaults

spec/api/v2/auth/auth_spec.rb
```diff
@@ -253,4 +253,50 @@
       end
     end
   end
+
+  describe 'test restrictions' do
```
ec: i will say, this should go in separate file.
Auth specs are too huge to read now, and its quite good to split them on files, e.q. file for rbac test, permissions test, auth endpoint itslef, restrictions

spec/api/v2/admin/restrictions_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+require 'spec_helper'
+describe API::V2::Admin::Restrictions do
+  include_context 'bearer authentication'
+
+  let!(:create_admin_permission) { create(:permission, role: 'admin', action: 'accept', verb: 'get') }
+  let!(:create_member_permission) { create(:permission, role: 'member') }
+
+  describe 'GET /api/v2/admin/restrictions' do
+    context 'successful response' do
+      it 'returns all restrictions' do
```
ec, line 17: missing pagination test

ec (approved):

PR #893, opened 2019-08-15
Add ability to configure uid prefix

ec:


config/initializers/user_config.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
```
mod: move to barong.rb with rdoc documentation for variables.
name it BARONG_UID_PREFIX="GRX"
validate it's only alpha (A-z) no more than 6 length

PR #903, opened 2019-08-22
Add session autorenew

dnfd:


config/initializers/barong.rb
```diff
@@ -38,6 +38,7 @@
 
 Barong::App.define do |config|
   config.set(:app_name, 'Barong')
+  config.set(:session_auto_renew, false)
```
ec: value should be in strings
lets stick to 'enabled' and 'disabled'

also var name should be prefixed with 'barong'

ec (changes requested):

ec:
This patch is conflicting with validations patch. Lets wait until we merge #904 first

ec (approved):

ec (approved):

mod (changes requested):
I don't agree with having a param
To me this should work well out of the box
What's the usecase ?
What are the API calls which triggers renew

More important is security like redis session and CSRF token on login

PR #904, opened 2019-08-22
Update Barong::App validations

dnfd:


config/initializers/cors.rb
```diff
@@ -1,5 +1,5 @@
 Barong::App.define do |config|
   config.set(:api_cors_origins, '*')
-  config.set(:api_cors_max_age, '3600')
-  config.set(:api_cors_allow_credentials, 'false')
+  config.set(:api_cors_max_age, '3600', regex: /^\d+$/)
+  config.set(:api_cors_allow_credentials, 'false', values: %w(false true))
```
mod: maybe we should convert as a boolean after?

dnfd: Yes, I thought about `type: :bool`

ec:
Thats an awesome improvement 
We should finish it as soon as possible)
It gives a lot more clearness and power for config store

@dnfd nice work!

ec:
@mod 
As we have moving so far with this, so maybe its time to prefix all envs with BARONG in the Barong::App class?
The idea is - we ask to provide env with a prefix, bur remove it in config.set

Then we will not have a duplicate in naming, like
`Barong::App.config.#{config_name}` and not `Barong::App.config.barong_#{config_name}`
But DevOps will be forced to set envs with `BARONG_CONFIG_NAME`

It will clarify a lot our deployments.
Although it's a breaking change for the current one (everything has its price)

nice catch from @dnfd

ec (approved):

PR #908, opened 2019-08-28
Change twilio API

chumaknadya:


app/models/phone.rb
```diff
@@ -37,31 +31,35 @@ def international(unsafe_phone)
...
+      Rails.logger.info("Sending code to #{phone.number}")
 
-      send_sms(number: phone.number,
-               content: Barong::App.config.sms_content_template.gsub(/{{code}}/, phone.code))
+      send_code(number: phone.number, channel: channel)
     end
 
-    def send_sms(number:, content:)
-      from_phone = Barong::App.config.twilio_phone_number
+    def send_code(number:, channel:)
```
ec, line 42: please extact to separate function

client = Barong::App.config.twilio_client
service_sid = Barong::App.config.twilio_service_sid

client.verify.services(service_sid)

ec (changes requested):

Gemfile
```diff
@@ -36,7 +36,7 @@ gem 'jwt', '~> 2.2'
 gem 'jwt-multisig', '~> 1.0'
 gem 'bunny'
 gem 'phonelib',     '~> 0.6.0'
-gem 'twilio-ruby',  '~> 5.6.0'
+gem 'twilio-ruby',  '~> 5.25.4'
```
ec, line 39: Why decreasing the Twilio version?
We should find a way to work with new versions.
We can't be stuck at this version forever

app/api/v2/resource/phones.rb
```diff
@@ -44,6 +44,11 @@ def validate_phone!(phone_number)
                    type: String,
                    allow_blank: false,
                    desc: 'Phone number with country code'
+          requires :channel,
+                   type: String,
+                   allow_blank: false,
+                   values: { value: %w(sms call), message: 'resource.phone.invalid_channel'},
```
ec: better we define a range of values in the model and use it everywhere, so we will not miss it anywhere, if any changes.
like `Phone::TWILIO_CHANNELS`
(think about naming)

app/api/v2/resource/phones.rb
```diff
@@ -55,7 +60,7 @@ def validate_phone!(phone_number)
           phone = current_user.phones.create(number: phone_number)
           code_error!(phone.errors.details, 422) if phone.errors.any?
 
-          Phone.send_confirmation_sms(phone)
+          Phone.send_confirmation(phone, declared_params[:channel])
           { message: 'Code was sent successfully' }
```
ec: Change the message, so we can understand what channel was used

app/api/v2/resource/phones.rb
```diff
@@ -72,16 +77,21 @@ def validate_phone!(phone_number)
                    type: String,
                    allow_blank: false,
                    desc: 'Phone number with country code'
+          requires :channel,
+                   type: String,
+                   allow_blank: false,
+                   values: { value: %w(sms call), message: 'resource.phone.invalid_channel'},
```
ec: the same.
Do you see? Already spreading hardcoded range in 2 files.

app/api/v2/resource/phones.rb
```diff
@@ -72,16 +77,21 @@ def validate_phone!(phone_number)
...
           validate_phone!(declared_params[:phone_number])
 
           phone_number = Phone.international(declared_params[:phone_number])
           phone = current_user.phones.find_by!(number: phone_number)
-          code_error!(phone.errors.details, 422) unless phone.regenerate_code
+          code_error!(phone.errors.details, 422) unless phone
 
-          Phone.send_confirmation_sms(phone)
+          Phone.send_confirmation(phone, declared_params[:channel])
           { message: 'Code was sent successfully' }
```
ec: message

app/api/v2/resource/phones.rb
```diff
@@ -72,16 +77,21 @@ def validate_phone!(phone_number)
...
+                   desc: 'The verification method to use'
         end
         post '/send_code' do
           declared_params = declared(params)
           validate_phone!(declared_params[:phone_number])
 
           phone_number = Phone.international(declared_params[:phone_number])
           phone = current_user.phones.find_by!(number: phone_number)
-          code_error!(phone.errors.details, 422) unless phone.regenerate_code
+          code_error!(phone.errors.details, 422) unless phone
```
ec: this doesn't make any sense now.
It was an error if the phone cant be updated. And now? This is invalid

app/api/v2/resource/phones.rb
```diff
@@ -107,12 +117,10 @@ def validate_phone!(phone_number)
...
           phone_number = Phone.international(declared_params[:phone_number])
-          phone = current_user.phones.find_by(number: phone_number,
-                                                 code: declared_params[:verification_code])
-
-          error!({ errors: ['resource.phone.verification_invalid'] }, 404) unless phone
+          verification = Phone.verify_code(number: phone_number, code: declared_params[:verification_code])
+          error!({ errors: ['resource.phone.verification_invalid'] }, 404) unless verification.status == 'approved'
 
-          phone.update(validated_at: Time.current)
+          current_user.phones.find_by(number: phone_number).update(validated_at: Time.current)
```
ec: the problem is, that now you don't check if the user trying to validate his own phone.
We should have an assignment `current_user.phones.find_by(number: phone_number)` before doing action with twilio

app/models/phone.rb
```diff
@@ -37,31 +31,35 @@ def international(unsafe_phone)
       parse(unsafe_phone).international(false)
     end
 
-    def send_confirmation_sms(phone)
-      Rails.logger.info("Sending SMS to #{phone.number}")
+    def send_confirmation(phone, channel)
+      Rails.logger.info("Sending code to #{phone.number}")
```
ec: the message should differ for channels. Better we now more info

app/models/phone.rb
```diff
@@ -37,31 +31,35 @@ def international(unsafe_phone)
...
 
-      send_sms(number: phone.number,
-               content: Barong::App.config.sms_content_template.gsub(/{{code}}/, phone.code))
+      send_code(number: phone.number, channel: channel)
     end
 
-    def send_sms(number:, content:)
-      from_phone = Barong::App.config.twilio_phone_number
+    def send_code(number:, channel:)
+      client = Barong::App.config.twilio_client
```
ec: prefix configs with barong_

config/initializers/phone.rb
```diff
@@ -1,24 +1,25 @@
...
     raise 'FATAL: Twilio setup is invalid'
   end
-  client = Barong::MockSMS.new(sid, token)
 else
   client = Twilio::REST::Client.new(sid, token)
+  service = client.verify.services.create(friendly_name: Barong::App.config.app_name) unless service_sid.present?
 end
 
-Barong::App.define { |c| c.set(:sms_sender, client) }
+Barong::App.set(:twilio_client, client) if client
```
ec: and what will happen if not `client`?
We will have errors in the project?

Better we think about this now

ec (changes requested):

Gemfile
ec: Uh, its 25 vs 6
Sorry)

mod:
looks more like a migration than an actual ability to switch.
I was thinking of making abstract class and config for choosing confirmation strategy

ec:
@mod are there any reasons for us not to actually `migrate` to this version of Twilio?
It is more smoothy, new, maintaining right now. 

Maybe just do a configuration file limiting channel to `SMS` only.

ec:
@mod just maintaining both is a bit difficult. And will spread garbage all over the project

ec:
will be reopened from branch soon

PR #916, opened 2019-09-04
Add system.session.create event with user and request_ip data

chumaknadya:


ec (approved):

spec/api/v2/identity/sessions_spec.rb
```diff
@@ -189,6 +189,20 @@
...
+    context 'event API behavior' do
+      before do
+        allow(EventAPI).to receive(:notify)
+      end
+
+      it 'receive system.session.create notify' do
+        post uri, params: { email: email, password: password }
+
+        expect(EventAPI).to have_received(:notify).with('system.session.create',
+          hash_including({ record: hash_including(user: anything, user_ip: anything, user_agent: anything) })
```
shal: Could we also test the logic?
Maybe we mock `user_agent` and `user_ip` and ensure, that it returns right stuff.

shal (changes requested):

shal:
Almost good, but I think we should add one more field with session time creation, then we will have something interesting.

I've tested it with postmaster, seems ok :+1:

PR #920, opened 2019-09-05
Add ability to restrict by geoip

dnfd:
- Country
  - Continent

### DIFF
--- a/lib/barong/app.rb
+++ b/lib/barong/app.rb
@@ -53,10 +53,17 @@ module Barong
         when :bool
           values!(key, value, %w(true false))
           return value == 'true'
-
         when :integer
           regex!(key, value,  /^\d+$/)
           return value
+        when :path
+          return Rails.root.join(value).tap { |p| path!(key, p) }
+        end
+      end
+
+      def path!(key, path)
+        unless File.exists?(path)
+          raise Error.new("#{key.to_s.upcase} path is invalid #{path.to_s}")
         end
       end
 
--- a/lib/barong/authorize.rb
+++ b/lib/barong/authorize.rb
@@ -69,15 +69,28 @@ module Barong
     end
 
     def validate_restrictions!
-      restrictions = Rails.cache.fetch('restrictions', expires_in: 5.minutes) do
-        Restriction.where(state: 'enabled').to_a
+      restrictions = Rails.cache.fetch('restrictions', expires_in: 5.minutes) { fetch_restrictions }
+
+      request_ip = @request.remote_ip
+
+      restrict! if restrictions['ip'].include?(request_ip)
+      restrict! if restrictions['ip_subnet'].any? { |r| IPAddr.new(r).include?(request_ip) }
+
+      scopes = Restriction::LOCATION_SCOPES
+      values = Barong::GeoIP.info(request_ip, *scopes)
+
+      Hash[scopes.zip(values)].each do |scope, value|
+        restrict! if value && restrictions[scope].any? { |r| r.casecmp?(value) }
       end
+    end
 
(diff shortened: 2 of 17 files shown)


config/initializers/geo_ip.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+Barong::App.set(:barong_maxminddb_path, '', type: :path)
```
ec: We try to move all the initializers like this in the 1 file.
Move to barong.rb

lib/barong/app.rb
```diff
@@ -57,6 +57,15 @@ def type!(key, value, options)
         when :integer
           regex!(key, value,  /^\d+$/)
           return value
+
```
ec: maybe we don't put an empty line?
i didn't find any tips about this style in rails.

lib/barong/authorize.rb
```diff
@@ -69,15 +69,28 @@ def api_key_owner
...
 
     def validate_restrictions!
-      restrictions = Rails.cache.fetch('restrictions', expires_in: 5.minutes) do
-        Restriction.where(state: 'enabled').to_a
+      restrictions = Rails.cache.fetch('restrictions', expires_in: 5.minutes) { fetch_restrictions }
+
+      request_ip = @request.remote_ip
+
+      restrict! if restrictions[:ip].include? request_ip
+      restrict! if restrictions[:ip_subnet].any? { |r| IPAddr.new(r).include?request_ip }
```
ec: better put (), like `include?(request_ip)` or `include? request_ip`

lib/barong/authorize.rb
```diff
@@ -69,15 +69,28 @@ def api_key_owner
...
+
+      scopes = %i[continent country city]
```
ec: maybe move scopes to model? this way we always keep it in one place and reuse

lib/barong/geo_ip.rb
```diff
@@ -0,0 +1,60 @@
...
+        @reader ||= MaxMind::DB.new(Barong::App.config.barong_maxminddb_path, mode: MaxMind::DB::MODE_MEMORY)
+      end
+
+      def fetch(record, key)
+        return unless record
+
+        case key.to_sym
+        when :country
+          return country(record)
+
```
ec: well, at least we use the same style everywhere

lib/barong/app.rb
```diff
@@ -57,6 +57,15 @@ def type!(key, value, options)
...
+        when :path
+          return Rails.root.join(value).tap { |p| path!(key, p) }
+        end
+      end
+
+      def path!(key, path)
```
ec, line 64: we also have path in barong, like path to seeds, path to label_state config.
Update them to :path type please

ec (approved):

lib/barong/authorize.rb
```diff
@@ -69,15 +69,28 @@ def api_key_owner
...
-      restrictions = Rails.cache.fetch('restrictions', expires_in: 5.minutes) do
-        Restriction.where(state: 'enabled').to_a
+      restrictions = Rails.cache.fetch('restrictions', expires_in: 5.minutes) { fetch_restrictions }
+
+      request_ip = @request.remote_ip
+
+      restrict! if restrictions[:ip].include?(request_ip)
+      restrict! if restrictions[:ip_subnet].any? { |r| IPAddr.new(r).include?(request_ip) }
+
+      scopes = Restriction::LOCATION_SCOPES.map(&:to_sym)
```
ec: why not just do a symbols array from the start?

ec (approved):

PR #921, opened 2019-09-06
Avoid no method error UTC for nil entities

ec:


app/api/v2/entities/activity.rb
```diff
@@ -3,7 +3,7 @@
 module API::V2
   module Entities
     class Activity < Grape::Entity
-      format_with(:iso_timestamp) { |d| d.utc.iso8601 }
+      format_with(:iso_timestamp) { |d| d&.utc&.iso8601 }
```
mod: so what will be format is d is nil ?

ec: it will be `nil`
But as soon, as a param is `nil` you dont need to format it.
It will just skip.

We already have the `d&.utc&.iso8601` in user entity and one more, and it works perfect

app/api/v2/entities/activity_with_user.rb
```diff
@@ -3,7 +3,7 @@
 module API::V2
   module Entities
     class ActivityWithUser < Grape::Entity
-      format_with(:iso_timestamp) { |d| d.utc.iso8601 }
+      format_with(:iso_timestamp) { |d| d&.utc&.iso8601 }
```
mod: mix in a helper
```ruby
def date_format(ts)
  if ts.nil?
....
  end
end
```

ec: is it required?
It looks very compact now. And safe operator fits well

PR #926, opened 2019-09-10
Add AliCloud Uploader

dnfd:


vshatravenko (approved):

ec (approved):

config/initializers/carrierwave.rb
```diff
@@ -10,6 +10,7 @@
   config.set(:storage_access_key, '')
   config.set(:storage_secret_key, '')
   config.set(:storage_region, '')
+  config.set(:uploader, UploadUploader)
```
ec: Make sure it's a string, as you can pass only strings in envs (or format it)

config/initializers/carrierwave.rb
```diff
@@ -31,6 +32,8 @@
     }
     config.fog_directory = Barong::App.config.storage_bucket_name
   elsif 'AliCloud'.casecmp?(Barong::App.config.storage_provider)
+    Barong::App.set(:uploader, AliUploader)
```
ec: So we reassign it even if env was another?
We should consider removing env and making it inside this if elsif

ec (changes requested):

config/initializers/carrierwave.rb
dnfd: This one will not be configured from env, I used Barong::App as a shared config store

dnfd: Maybe I'll disable env for this later

vshatravenko (approved):

PR #939, opened 2019-09-23
CSRF protection

ec:


app/api/v2/entities/user_with_full_info.rb
```diff
@@ -24,6 +25,16 @@ class UserWithFullInfo < API::V2::Entities::Base
...
         expose :profile, using: Entities::Profile
         # activities, as sensitive and potentialy too big data should be queried separately
+
+        private
+
+        def secret
+          # csrf protection token
+          secret = SecureRandom.base64(32)
+
+          Rails.cache.write(object.uid, secret)
```
calj: we need to make sure this cache expire as the same date of the user session.

calj:
We need somehow to bypass this mechanism for queries using API keys.

app/api/v2/entities/user_with_full_info.rb
mod: I want to switch to redis session in the same patch and store the token at this same location

app/controllers/authorize_controller.rb
```diff
@@ -35,4 +37,13 @@ def access_error!(text, code)
     response.status = code
     response.body = { 'errors': [text] }.to_json
   end
+
+  def csrf_token_validate?
+    if request.get? || request.head? || (request.headers.key?('X-CSRF-Token') &&
+       request.headers['X-CSRF-Token'] == Rails.cache.read(session[:uid]))
```
mod: That is a very confusing redis key

app/api/v2/entities/user_with_full_info.rb
mod: This isn't a secret but a csrf token, also I would put it in entity is there a way we can show it only in login (session) controller ? Maybe

session {
  token: ...,
  expire: server time,
  user: entity
}

Also it would be nice if we protect sessions by IP
So we store session server side, among a token, uid, user-ip, user-agent

If token provided, user-ip, user-agent has changed we consider session invalid

mod (changes requested):

mod (changes requested):
We need to store in redis session, uid, user-ip, user-agent and token

And when validating active session, if user-agent, user-ip, or token provided is different than what was stored in the session we destroy the session and logout.

mod:
On session create, we save user-ip, user-agent, and generate a random token and save it

We create session in server side with a random session id, which must be stored in cookie and nothing else.

When we validate current session we read again from query ip, uagent, token (from header) and compare

app/api/v2/entities/user_with_full_info.rb
mod: This is called a token not a secret

ec:
duplicated #958

PR #948, opened 2019-10-03
Server side sessions implementation

ec:


ec:
We have choose another way of implementation in #949

PR #955, opened 2019-10-15
Protect superadmin against changes from non-superadmin users

ec:


mod:
I see you protect labels too ?
Why not using cancan here ?

mod (approved):

PR #964, opened 2019-10-24
Reorganize docs && add password_hashing.md

ec:


mod (changes requested):
can you rebase and use widdershins to generate api docs
this is the compatible tool with SDK

PR #965, opened 2019-10-28
Add cloudflare middleware to improve remote_ip

ec:


lib/barong/cloudflare_middleware.rb
```diff
@@ -0,0 +1,58 @@
...
+                    IPAddr.new(matcher)
+                  else
+                    raise ArgumentError, "Expected String, IPAddr or Regexp but found #{matcher.class} #{matcher.inspect}"
+                  end
+    end
+  end
+
+  def call(env)
+    @env = env
+    ips = ips_from('HTTP_X_FORWARDED_FOR')
```
jloup, line 30: You call ips_from on 'HTTP_X_FORWARDED_FOR' string, not header's content.

PR #970, opened 2019-11-06
Feature: add configuration and troubleshooting docs

chumaknadya:


docs/troubleshooting.md
```diff
@@ -0,0 +1,19 @@
...
+1. Check list of your permissions on barong seeds.yml or tower admin panel
+2. If there are no such permission just add new permission on tower admin panel
+
+- `401 authz.invalid_permission` after login
+1. Check your permissions on barong rails console 
+```ruby
+irb(main)013:0> Permission.all
+irb(main)013:0> Rails.cache.read('permissions')
+```
+2. If it is wrong/empty just update it using following command
```
ec: its not an update.
I guess we need to explain mechanism a bit, as its a bit not clear

docs/configuration.md
```diff
@@ -0,0 +1,113 @@
...
+
+---
+
+## State configuration
+
+We can customize barong configuration as we want
+
+1. For user activation we just need to have verified email label in example below. You can put  more labels to create your own rules for user activation
+2. For example, if you want to ban your user you just need to put ban and fraud labels on tower admin panel. For sure you can customize this case too and put change or add label names in barong.yml
+3. For document verification we use following document types, which can be changed as you wish
```
ec: For document verification we use, as standard - following document types.
But you can configure available document types by changing or extending existing list. This way we keep an opportunity to support any custom KYC services, logic, etc

docs/configuration.md
```diff
@@ -0,0 +1,113 @@
...
+
+reCAPTCHA is a CAPTCHA-like system designed to establish that a computer user is human (normally in order to protect websites from bots) and, at the same time, assist in the digitization of books or improve machine learning. 
+
+You can learn more about how to create ``RECAPTCHA_SECRET_KEY``, ``RECAPTCHA_SITE_KEY`` in this [article](https://writeup.xyz/how-to/google-recaptcha-v2-tutorial-3125/)
+
+---
+
+## Blacklist/Whitelist configuration
+
+You need to put whitelisted (public) routes for pass object and blacklisted routes for block in authz_rules.yml
```
ec, line 72: `Pass` routes will never be checked by AuthZ endpoint and will be available without session requirement.
On `Block` routes user always will get 401, it doesn't depend on a session / role / ip / etc

docs/configuration.md
```diff
@@ -0,0 +1,113 @@
...
+- `ALIYUN_REGION_ID`
+- `ALIYUN_OSS_ENDPOINT`
+
+  [Learn more about how to create Access key](https://www.alibabacloud.com/help/doc-detail/53045.htm)
+
+---
+
+
+
+## Recaptcha configuration
```
ec, line 60: will we describe about NO captcha and pazzle captcha?

docs/configuration.md
```diff
@@ -0,0 +1,113 @@
...
+We have ability to set twilio with 3 different ways
+1. ```BARONG_PHONE_VERIFICATION == "twilio_sms"```
+  If you choose phone verification as twilio sms we will use send_sms [API call](https://www.twilio.com/docs/sms/send-messages)
+  Also you can add your own template for sms using `BARONG_SMS_CONTENT_TEMPLATE`
+2. ```BARONG_PHONE_VERIFICATION == "twilio_verify"```
+  In this case we will use twilio verify [API call](https://www.twilio.com/docs/verify/api)
+  There are a lot of benefits of using this API like you can send your sms via voice
+  One verification service can be used to send multiple verification tokens, it is not necessary to create a new service each time, so you can set ```BARONG_TWILIO_SERVICE_SID``` at once
+3. ```BARONG_PHONE_VERIFICATION == "mock"```
+  Use this case only for local testing
```
ec: With this type of verification all the numbers will be accepted and validated as a right code for any given number

docs/configuration.md
```diff
@@ -0,0 +1,113 @@
...
+- `BARONG_TWILIO_SERVICE_SID`, which acts as a twilio password
+- `BARONG_TWILIO_PHONE_NUMBER`, virtual phone numbers which will give you instant access to local, national, mobile, and toll-free phone numbers
+
+We have ability to set twilio with 3 different ways
+1. ```BARONG_PHONE_VERIFICATION == "twilio_sms"```
+  If you choose phone verification as twilio sms we will use send_sms [API call](https://www.twilio.com/docs/sms/send-messages)
+  Also you can add your own template for sms using `BARONG_SMS_CONTENT_TEMPLATE`
+2. ```BARONG_PHONE_VERIFICATION == "twilio_verify"```
+  In this case we will use twilio verify [API call](https://www.twilio.com/docs/verify/api)
+  There are a lot of benefits of using this API like you can send your sms via voice
```
ec: `send your sms via voice`
doesnt seem legit

ec (changes requested):

ec (approved):

PR #981, opened 2019-11-25
Feature: add DataStorages

ec:


db/schema.rb
```diff
@@ -38,6 +38,15 @@
     t.index ["user_id"], name: "index_apikeys_on_user_id"
   end
 
+  create_table "data_storages", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
+    t.bigint "user_id", null: false, unsigned: true
+    t.string "title", limit: 64, null: false
+    t.text "data", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["user_id", "title"], name: "index_data_storages_on_user_id_and_title"
```
calj: this must be a unique index

app/api/v2/resource/data_storage.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module API
+  module V2
+    module Resource
+      # data storage C functionality
+      class DataStorage < Grape::API
+        resource :data_storage do
+          desc 'Create an api key',
```
calj: Description is wrong here

config/barong.yml
```diff
@@ -16,3 +16,6 @@ document_types:
   - Utility Bill
   - Residental
   - Institutional
+data_storage_titles:
```
calj: I would set the default value as empty array here to prevent accidentally activating user storage.

calj: I would name it `user_storage_titles`

config/initializers/data_storage_titles.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# whitelisted data storage titles definitions
+class DataStorageTitles
+  class << self
+    def list
+      @list ||= YAML.load_file(Barong::App.config.barong_config)['data_storage_titles']
```
calj: default it to [] if unset

ec: you mean if key `data_storage_titles` dont exist at all?

config/barong.yml
ec: about `user_storage` prefix dont you think its a bit unclear?
user_storage sounds like a storage of users?

ec: okay. but i will keep possible options as a comment to show possible functionality

calj: user_personal_storage_titles_whitelist ?

config/initializers/data_storage_titles.rb
calj: yes to not brake on upgrade of barong if we don't update the configuration

PR #986, opened 2019-12-18
Feature: introduce CSRF protection

ec:
Simple and effective CSRF protection overview and protection principles https://www.linkedin.com/pulse/how-protect-your-app-from-csrf-artem-linetskyi by @alinetskyi

### DIFF
--- a/app/api/v2/identity/users.rb
+++ b/app/api/v2/identity/users.rb
@@ -60,9 +60,9 @@ module API::V2
           activity_record(user: user.id, action: 'signup', result: 'succeed', topic: 'account')
 
           publish_confirmation(user, language, Barong::App.config.domain)
-          open_session(user)
+          csrf_token = open_session(user)
 
-          present user, with: API::V2::Entities::UserWithFullInfo
+          present user, with: API::V2::Entities::UserWithFullInfo, csrf_token: csrf_token
           status 201
         end
 
@@ -126,7 +126,8 @@ module API::V2
             end
 
             current_user.labels.create!(key: 'email', value: 'verified', scope: 'private') if token_uniq?(payload[:jti])
-            open_session(current_user)
+
+            csrf_token = open_session(current_user)
 
             EventAPI.notify('system.user.email.confirmed',
                             record: {
@@ -135,7 +136,7 @@ module API::V2
                               domain: Barong::App.config.domain
                             })
 
-            present current_user, with: API::V2::Entities::UserWithFullInfo
+            present current_user, with: API::V2::Entities::UserWithFullInfo, csrf_token: csrf_token
             status 201
           end
         end
--- a/lib/barong/authorize.rb
+++ b/lib/barong/authorize.rb
@@ -5,6 +5,7 @@ require 'barong/activity_logger'
 module Barong
   # AuthZ functionality
   class Authorize
+    STATE_CHANGING_VERBS = %w[POST PUT PATCH DELETE TRACE].freeze
     # Custom Error class to support error status and message
     class AuthError < StandardError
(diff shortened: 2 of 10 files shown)


lib/barong/authorize.rb
```diff
@@ -35,6 +35,8 @@ def auth
 
     # cookies validations
     def cookie_owner
+      validate_csrf! if Barong::App.config.barong_csrf_protection
```
alinetskyi: I believe here we validate every single request, regardless of whether it's GET or POST etc.
However according to the official document (https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#javascript-guidance-for-auto-inclusion-of-csrf-tokens-as-an-ajax-request-header)

POST, PUT, PATCH, DELETE and TRACE verbs, being state changing verbs, should have a CSRF token attached to the request.

alinetskyi (changes requested):

lib/barong/authorize.rb
ysv: good point @alinetskyi

ysv (changes requested):
1. I advice we take a look into default ActionController implementation of CSRF `ActionController::RequestForgeryProtection` and either we use it or follow the same principle
2. https://medium.com/rubyinside/a-deep-dive-into-csrf-protection-in-rails-19fa0a42c0ef

ysv (approved):

lib/barong/authorize.rb
```diff
@@ -35,6 +36,10 @@ def auth
 
     # cookies validations
     def cookie_owner
+      if Barong::App.config.barong_csrf_protection && @request.env['REQUEST_METHOD'].in?(STATE_CHANGING_VERBS)
```
mod: I'd move the if inside validate_token

mod (changes requested):
Looks good
Just need rebase and some documentation of the param.

lib/barong/authorize.rb
ec: yes, this way we save 2 lines)
done

app/api/v2/identity/users.rb
```diff
@@ -60,9 +60,10 @@ def parse_refid!
           activity_record(user: user.id, action: 'signup', result: 'succeed', topic: 'account')
 
           publish_confirmation(user, language, Barong::App.config.domain)
-          open_session(user)
+          token = SecureRandom.hex(10)
```
mod: this must be in open_session, to me it's part of this function
you repeat it 3 times in the code, if we want to change the SecureRandom we need to do it 3 times
size(10) can be a param or a DEFAULT

spec/api/v2/auth/restriction_spec.rb
```diff
@@ -4,7 +4,9 @@
 
 describe '/api/v2/auth functionality test' do
   include_context 'geoip mock'
-
+  before do
+    allow_any_instance_of(Barong::Authorize).to receive(:validate_csrf!).and_return(true)
```
mod: maybe we disable CRSF for tests instead ?

app/api/v2/identity/users.rb
ec: I need to have it to give to the entity as a param.
If I will move it in the function, then I will need function to return value. (Which is already misunderstanding of open_session aim)
Save value in a var and give to entity .

It will become ugly

mod (changes requested):
I would be in facvor of having a test file for CSRF, and disable it for the other test files

mod (approved):

PR #1003, opened 2020-01-10
Enhauncement: allow to set min_entropy and use_dictionary via env

akushniruk:


ec:
We will switch to use https://github.com/dropbox/zxcvbn

PR #1006, opened 2020-01-20
Feature: rework password validation policy

ec:


Gemfile
```diff
@@ -65,7 +65,7 @@ gem 'bump'
 # Use gem to verify recatpcha on server side
 gem 'recaptcha'
 # Password validators
-gem 'strong_password', '~> 0.0.6'
+gem 'zxcvbn-js', require: 'zxcvbn'
```
mod: Why don't you use the pure ruby version ?

config/initializers/barong.rb
```diff
@@ -35,11 +35,13 @@
 
 Barong::App.define do |config|
   # General configuration ---------------------------------------------
-  # https://github.com/openware/barong/blob/master/docs/general/env_configuration.md#general-configuration
+  # https://github.com/openware/barong/blob/master/docs/configuration.md#general-configuration
 
   config.set(:app_name, 'Barong')
   config.set(:domain, 'openware.com')
   config.set(:uid_prefix, 'ID', regex: /^[A-z]{2,6}$/)
+  config.set(:password_min_score, '2', type: :integer)
```
mod: Default min score must be 3

config/initializers/barong.rb
```diff
@@ -35,11 +35,13 @@
...
+  config.set(:password_min_length, '5', type: :integer)
```
mod: Default min length should be 8

mod (changes requested):
Looks good, but we should use the pure Ruby version execjs is really a bad practice

PR #1015, opened 2020-02-17
Feature: Add ability to support multiple profiles

ec:


app/api/v2/admin/profiles.rb
```diff
@@ -36,14 +36,16 @@ class Profiles < Grape::API
...
             optional :city, type: String
+            optional :state, type: String
             optional :country, type: String
             optional :metadata, type: String, desc: 'Any additional key: value pairs in json string format'
           end
 
           put do
-            target_profile = User.find_by(uid: params[:uid])&.profile
-            return error!({ errors: ['admin.profiles.doesnt_exist'] }, 404) if target_profile.nil?
+            target_profile = User.find_by(uid: params[:uid])&.profiles&.find_by(state: 'drafted')
```
mod: better make model methods we can understand like 
latest_draft
latest_approved
something like that
avoid 1 liners it's evil

chumaknadya (approved):

PR #1017, opened 2020-02-18
Add ability to configure upload-related configuration

ec:


app/uploaders/upload_uploader.rb
```diff
@@ -17,20 +17,22 @@ def store_dir
   # Add a white list of extensions which are allowed to be uploaded.
   # For images you might use something like this:
   def extension_whitelist
+    Barong::App.config.barong_upload_extension_whitelist
     %w[jpg jpeg png pdf]
```
mod: I think this is wrong

PR #1021, opened 2020-02-19
Enhancement: security updates

ec:
* Make default protection with captcha on email send endpoints
* Allow to configurate endpoints you want to protect with captcha
* Prevent user enumeration on identity/* endpoints

### DIFF
--- a/app/api/v2/identity/sessions.rb
+++ b/app/api/v2/identity/sessions.rb
@@ -23,14 +23,11 @@ module API::V2
                    desc: 'Code from Google Authenticator'
         end
         post do
+          verify_captcha!(response: params['captcha_response'], endpoint: 'session_create')
+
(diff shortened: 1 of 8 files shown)


pkucherenk0:
1. Login flow with 2FA enabled and captcha will be difficult as if the user will have 2FA enabled he will be forced to pass it twice on the initial attempt to log in, where we get a response that the user has 2FA and on a page where the user will enter OTP.
(Most probably frontend will have to figure out the flow)

2. On resend confirmation email endpoint if the user exists the process will be fine, but if the 
Selection_404
user does not exist endpoint will respond with `201` and ignore the captcha.
Selection_403

3. On reset password same problem as with resend confirmation email and have to be configured as `reset_password`, not as `password_reset`

PR #1037, opened 2020-03-19
Fix: Remove log level WARN for response status code 5xx

chumaknadya:


ec (approved):

mod (changes requested):
This is just wrong
To enable debug logs in production we will need a special FLAG which may already exists in RAILS

ec:
As soon as no action proceed on the topic, im closing PR. Please, let me know if we need to rework it.

PR #1038, opened 2020-03-20
Feature: redeploy on master.devkube.com

dpatsora:


ec (approved):

ec:
@dpatsora maybe we can do such things configurable? then we dont need to edit it each time

vshatravenko (approved):

vshatravenko:
Yeah would be nice to use secrets only

PR #1041, opened 2020-03-23
Feature/email template

Kohelbekker:


ec (changes requested):
Why do you have 2 commits here?
I assume you have some rebase problems

PR #1044, opened 2020-03-30
Enhancement: Add created_at and updated_at to profile entities

chumaknadya:


ec (approved):

mod:
we should import this with merging of 2-4-stable in master

PR #1050, opened 2020-04-13
Feature: PUT resource/users/me data field

ec:


mod (changes requested):
I do not understand the context
Can you remind me the usecase

PR #1052, opened 2020-04-14
Feature: admin endpoint to retrieve user api keys

ec:


spec/api/v2/admin/api_keys_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        uid: test_user.uid
+      }
+    end
+    let(:do_request) { get "/api/v2/admin/api_keys", params: params, headers: auth_header }
+
+    context 'successful' do
+      it 'Returns api keys for selected account' do
+        do_request
+        expect(response.status).to eq(200)
+        expect(json_body.first).to include(expected_fields)
```
calj, line 37: can you ensure the secret is absent from the response here

calj (changes requested):
please improve test to ensure secrets are absents from the response

spec/api/v2/admin/api_keys_spec.rb
ec: yes it is. But we will need to change the flow once we get merged new Vault policies, as Maksim has changed the way it work for secret key for API Keys.

I will add it in the test, so it will fail on Vault rebased patch

PR #1054, opened 2020-04-17
Enhancement: allow only latest requested token on reset_pass

ec:


app/api/v2/identity/users.rb
```diff
@@ -176,7 +176,10 @@ def parse_refid!
 
             return status 201 if current_user.nil?
 
-            token = codec.encode(sub: 'reset', email: params[:email], uid: current_user.uid)
+            reset_pass_id = SecureRandom.hex(10)
+            token = codec.encode(sub: 'reset', email: params[:email], uid: current_user.uid, reset_pass_id: reset_pass_id)
+            # save reset_password_id in cache to validate as latest requested
+            Rails.cache.write("reset_password_#{params[:email]}", reset_pass_id)
```
mod: maybe we should do it the devise way, storing this id in DB ?
cache expire and disappear maybe this should be in redis for the duration of the validity of the jwt ?

ec: @mod no point to store in the database, as soon as we store only the last VALID identifier of the token requested. so it always be only 1 cache record per user.
Yes we can do `expires_in: token_lifetime`, good idea

app/api/v2/identity/users.rb
```diff
@@ -176,7 +176,10 @@ def parse_refid!
 
             return status 201 if current_user.nil?
 
-            token = codec.encode(sub: 'reset', email: params[:email], uid: current_user.uid)
+            reset_pass_id = SecureRandom.hex(10)
```
mod: ideally rename reset_token

mod (approved):

PR #1055, opened 2020-04-17
Enhancement: allow to configure X-Frame-Options (DENY as default)

ec:


ec:
frozen, can be done BEFORE server

PR #1057, opened 2020-04-22
Feature: support TRUE_CLIENT_IP header as IP source

ec:


app/api/v2/utils.rb
```diff
@@ -3,10 +3,18 @@
...
     def remote_ip
-      ip_string = env['action_dispatch.remote_ip'].to_s
-      Rails.logger.debug "User login IP address: #{ip_string}"
+      # default behaviour, IP from HTTP_X_FORWARDED_FOR
+      ip = env['action_dispatch.remote_ip'].to_s
 
-      ip_string
+      if Barong::App.config.allow_true_client_ip
+        # custom header that contains only client IP
+        true_client_ip = request.env['HTTP_TRUE_CLIENT_IP'].to_s
```
mod, line 11: are you sure we should have this option,
can't we just define an order of request env which we will always follow ?

if not maybe we do a full support to akamai, we the flag should be

BARONG_WAF=cloudflare/akamai

ec: @mod first of all, not only akamai uses true-client-ip, then, its already a full support of akamai.
And, this implementation is exactly the way akamai suggest it to be.

mod: so just remove if Barong::App.config.allow_true_client_ip
let's always consider this header and probably in the highest priority

PR #1058, opened 2020-04-27
Enhancement: remove api_key signature blacklisting

ec:


mod (changes requested):
in addition we need some leeway for the future timestamp
or remove this check, I think we can allow expire_time / 2 in the future ?

ec:
> in addition we need some leeway for the future timestamp
> or remove this check, I think we can allow expire_time / 2 in the future ?

@mod eventually, it will increase our "window", which can work, but i dont see the sense.

lib/barong/authorize.rb
```diff
@@ -82,18 +82,11 @@ def find_ip
...
-      end
       # validate that nonce is a positive integer
       error!({ errors: ['authz.nonce_not_valid_timestamp'] }, 401) if api_key_params[:nonce].to_i <= 0
       # timestamp_window is a difference between server_time and nonce creation time
-      nonce_timestamp_window = (Time.now.to_f * 1000).to_i - api_key_params[:nonce].to_i
-      # (server_time - nonce) should be positive
-      error!({ errors: ['authz.nonce_from_future'] }, 401) if nonce_timestamp_window < 0
-      # (server_time - nonce) should not be more than nonce lifetime
-      error!({ errors: ['authz.nonce_expired'] }, 401) if nonce_timestamp_window >= Barong::App.config.apikey_nonce_lifetime
+      api_key_params[:nonce].to_s.length == 13 ? validate_milisecond_nonce! : validate_second_nonce!
```
mod: no, only seconds or millisecond, I don't want to support both.
stick to milliseconds

PR #1061, opened 2020-04-30
Feature: introduce category of restrictions, restrictions as firewall

ec:


app/api/v2/identity/users.rb
```diff
@@ -17,6 +17,32 @@ def parse_refid!
 
       desc 'User related routes'
       resource :users do
+        desc 'Creates new user',
```
mnaichuk: Invalid desc

mod (approved):

PR #1062, opened 2020-05-04
Update Barong vault policies

mnaichuk:


config/initializers/barong.rb
```diff
@@ -69,6 +69,7 @@
   config.set(:event_api_rabbitmq_password, 'guest')
   config.set(:vault_address, 'http://localhost:8200')
   config.set(:vault_token, 'changeme')
+  config.set(:vault_app_name, 'barong')
```
ec, line 73: please document the env in docs/configuration.md

ec: and we already have `config.set(:app_name, 'Barong')` maybe we can reuse it?

mnaichuk: I guess vault_app_name and app_name can be different.

ec (approved):

ec (approved):

ec (approved):

PR #1066, opened 2020-05-07
Refactor: remove nonsense failed login activity on empty otp

ec:


caiovilas:
The message for failed login for missing OTP is actually quite important. If a user password is leaked and somebody tries to access an account without OTP, it is important to have the failed login attempt logged for legal reasons and for informing the user.
Why is it being removed in the first place?

calj:
It should still log invalid OTP.
This change should just stop logging error with empty OTP.

ec:
Yes @calj is right. @caiovilas  activity will be recorded if any invalid otp will be submitted. This check had no sense only because due to the software design, we don't show otp field by default, so every login with 2fa turned on always started with a failed activity.

crackfoo:
> Yes @calj is right. @caiovilas activity will be recorded if any invalid otp will be submitted. This check had no sense only because due to the software design, we don't show otp field by default, so every login with 2fa turned on always started with a failed activity.

how can we display the OTP field?

calj:
Hi @crackfoo,

You can't know if the user enabled the OTP before he logs in.
You can make the login in one step and ask for the OTP straight, mark it as optional so the user would fill it if needed. On [BaseApp](https://github.com/openware/baseapp) we made this process in two steps:
1. The user provides email and password, if the otp is mandatory barong will return a 401
2. We request additionally the OTP only if we got a 401 on the first step

This is more user friendly so he doesn't need to remember if the OTP is enable on his account.

More info about the session endpoint here: [OpenDAX API](https://www.openware.com/sdk/2.6/docs/barong/api/barong-user-api-v2.html#api-v2-barong-identity-sessions)

crackfoo:
@calj I see that the process has been somewhat fixed in baseapp, but I can only get the login to work as expected using the absolute latest quay.io/openware/baseapp:2.6.0-2d57ca8 when then in-turn causes problems in user's wallet page.

Is there a known working combination of baseapp/barong versions that work well togther?

I'm currently using:

  peatio: quay.io/openware/peatio:2.5.46
  barong: quay.io/openware/barong:2.5.20
  frontend: quay.io/openware/baseapp:2.3.29
  tower: quay.io/openware/tower:2.5.27
  postmaster: quay.io/openware/postmaster:2.3.4
  rango: quay.io/openware/rango:2.5.0
 
Thanks

crackfoo:
I've got it fixed and working as expected doing a custom build using baseapp 2.4.1

PR #1090, opened 2020-06-02
Feature: post admin/profiles, 2-admin profile approval

ec:


app/api/v2/admin/profiles.rb
```diff
@@ -49,12 +49,48 @@ class Profiles < Grape::API
               error!({ errors: ['admin.profiles.superadmin_change'] }, 422)
             end
 
+            if target_profile.author && target_profile.author == current_user.uid
```
dnfd: Could we add a configurable list of roles for users who can (approve) update profiles? F.e. if I want to allow superadmins only to approve profiles. :smiley_cat:

ec: Technically, yes, but if there is a point for it? You can manage it through permissions

dnfd (approved):

calj:
can you write a short documentation about this feature?

PR #1098, opened 2020-06-09
Add ability for admin to create comments

dnfd:


ec (approved):

app/api/v2/admin/users.rb
```diff
@@ -512,6 +512,89 @@ def search(field, value)
...
+              { code: 401, message: 'Invalid bearer token' }
+            ]
+            params do
+              requires :uid,
+                       type: String,
+                       allow_blank: false,
+                       desc: 'user uniq id'
+              requires :title,
+                       type: String,
+                       allow_blank: false,
```
calj, line 538: let limit the length here to the same as in db for title and data

app/api/v2/admin/users.rb
```diff
@@ -512,6 +512,89 @@ def search(field, value)
...
+            desc "Edit user's comment",
+            security: [{ "BearerToken": [] }],
+            failure: [
+              { code: 401, message: 'Invalid bearer token' }
+            ]
+            params do
+              requires :id,
+                       type: Integer,
+                       desc: 'comment uniq id'
+              optional :title,
```
calj, line 571: let limit the length here to the same as in db for title and data

PR #1106, opened 2020-06-25
Enhancement: Use appropriate words

mikoim:
The technology industry has started to replace problematic racist terminology with more appropriate and precise. It's time to shift from naming conventions, which cause misunderstanding.


ec:
Thanks for the contribution, we already have plans for changing namings to appropriate words. We will keep you posted about the progress on our side

PR #1108, opened 2020-06-30
Enhancement: Add ordering to admin users, api_keys and resources api_keys endpoint

dpatsora:
…_keys endpoint


ec:
We should check whats wrong with guardrails. Maybe just repush, so it will re-trigger

ec (approved):

PR #1109, opened 2020-06-30
check 2FA code with management API

sundanny26:
It give other applications the possibility to use 2fa code used by barong. So you can add another layer of security.


ec:
Hi @sundanny26 ! Its a great idea, but this will come soon as a part of another patch. Follow the updates, i will ping you there once it will be merged

PR #1114, opened 2020-07-09
Feature: ability to customize email template logo

Kohelbekker:
Feature requested by @vshatravenko, should be implemented along with https://github.com/openware/opendax/pull/184

### DIFF
--- a/app/mailers/postmaster.rb
+++ b/app/mailers/postmaster.rb
@@ -7,6 +7,7 @@ class Postmaster < ApplicationMailer
     @record  = params[:record]
     @changes = params[:changes]
     @user    = params[:user]
+    @logo    = params[:logo]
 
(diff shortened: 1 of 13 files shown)


bin/mailer
```diff
@@ -20,7 +20,7 @@ class Mailer
     c.desc 'Run processing email events'
     c.action do |global_options, _options, _args|
       config = YAML.load_file(global_options[:config]).deep_symbolize_keys
-      EventMailer.call(config[:events], config[:exchanges], config[:keychain])
+      EventMailer.call(config[:events], config[:exchanges], config[:keychain], config[:mailer_logo])
```
ec: Why not just using an ENV? This looks too complicated for no reason for me

ec (changes requested):

ec (approved):

PR #1115, opened 2020-07-14
Feature: Add redis cluster support

mnaichuk:


config/initializers/barong.rb
```diff
@@ -71,7 +71,9 @@
   config.set(:event_api_rabbitmq_password, 'guest')
   config.set(:vault_address, 'http://localhost:8200')
   config.set(:vault_token, 'changeme')
+  config.set(:vault_cluster, 'false')
```
ec: Please, mark as `type: bool`, this way config store will transfer it to boolean and you will be able to use it as a bool in the code

ec (changes requested):

ec (approved):

PR #1117, opened 2020-07-16
Fix: Fix seed_restrictions function

akulakovaa:


ec:
i guess we will need more here, as its not the only problem. Ill squash in 1 patch

PR #1119, opened 2020-07-17
Feature: add filters for /resource/users/activity/all endpoint

vpetrusenko:


app/api/v2/resource/users.rb
```diff
@@ -68,12 +68,29 @@ def verify_otp!
                    type: String,
                    allow_blank: { value: false, message: 'resource.user.empty_topic' },
                    desc: 'Topic of user activity. Allowed: [all, password, session, otp]'
+          optional :time_from,
+                   type: { value: Integer, message: 'resource.user.non_integer_time_from' },
+                   allow_blank: { value: false, message: 'resource.user.empty_time_from' },
+                   desc: "An integer represents the seconds elapsed since Unix epoch."\
+                         "If set, only orders created after the time will be returned."
```
ec: `only orders` - what orders XD

Seems like just a copypaste problem.
Please, check all of the descs

spec/api/v2/resource/users_spec.rb
```diff
@@ -167,6 +167,25 @@
...
+      expect(activities.length).to eq(4)
+    end
+    it 'sorts user activities by time range' do
+      4.times { create(:activity, user: test_user, result: 'succeed') }
+      get '/api/v2/resource/users/activity/all', params: { time_from: 5.hours.ago.to_i, time_to: 1.hours.ago.to_i }, headers: auth_header
+      activities = JSON.parse(response.body)
+      expect(activities.length).to eq(1)
+    end
+
+    it 'sorts user activities by time range' do
```
ec, line 183: i believe this tests doesn't show the use case, as you have only 4 activity in DB, and you fetch all 4.

Create 4 activity older then 5 hours, and 4 fresh. And check if filtering works correctly, dividing them

ec (changes requested):

ec (approved):

PR #1137, opened 2020-08-19
Introduce service accounts

ec:


app/api/v2/management/service_accounts.rb
```diff
@@ -0,0 +1,87 @@
...
+          declared_params = declared(params, include_missing: false)
+
+          service_acc = ServiceAccount.find_by!(declared_params)
+          error!('Service account doesnt exist', 422) unless service_acc
+
+          present service_acc, with: API::V2::Entities::ServiceAccounts
+          status 200
+        end
+
+        desc 'Get service_accounts as a paginated collection' do
```
dpatsora, line 31: Maybe it will be good to add an ability, to filter service_accounts by UID of User to which these account belongs?

app/api/v2/management/service_accounts.rb
```diff
@@ -0,0 +1,87 @@
...
+          service_accs.tap { |q| present paginate(q), with: API::V2::Entities::ServiceAccounts }
+          status 200
+        end
+
+        desc 'Create service_account' do
+          @settings[:scope] = :write_service_accounts
+          success API::V2::Entities::ServiceAccounts
+        end
+        params do
+          requires :service_account_uid, type: String, allow_blank: false, desc: 'service_account uid'
```
dpatsora: Why do we pass UID here?
Maybe you will generate it inside Barong ( same as we do [here](https://github.com/openware/barong/blob/a97b181937dcf01f510ba1e7947dff262541e99b/app/api/v2/management/users.rb#L90))

app/models/document.rb
```diff
@@ -59,13 +59,16 @@ def doc_expire_not_in_the_past
...
+#  user_id       :bigint           unsigned, not null
+#  upload        :string(255)
+#  doc_type      :string(255)
+#  doc_number    :string(255)
+#  doc_expire    :date
+#  identificator :string(255)
+#  metadata      :text(65535)
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#  doc_issue     :date
```
dpatsora: How it relates to this PR?

app/models/profile.rb
```diff
@@ -3,19 +3,21 @@
...
-#  dob        :date
-#  address    :string(255)
-#  postcode   :string(255)
-#  city       :string(255)
-#  country    :string(255)
-#  state      :integer          default("drafted"), unsigned
-#  metadata   :text(65535)
-#  created_at :datetime         not null
-#  updated_at :datetime         not null
+#  id           :bigint           not null, primary key
```
dpatsora: Why it relates to this PR?

app/models/user.rb
```diff
@@ -57,7 +58,7 @@ def superadmin?
   def role_exists
     return if Permission.pluck(:role).include?(role)
 
-    errors.add(:role, 'doesnt_exist')
+    # errors.add(:role, 'doesnt_exist')
```
dpatsora: Why do we commented this?

ec: Thats a mistake, i did when tested locally. good catch)

app/models/profile.rb
ec: We havent updated annotates for a long time (its not automatic in barong). As soon as i have added new model - i've re-runed an annotations. You can see there is no difference in schema but only in annotations in models)

app/models/document.rb
ec: We havent updated annotates for a long time (its not automatic in barong). As soon as i have added new model - i've re-runed an annotations. You can see there is no difference in schema but only in annotations in models)

app/api/v2/management/service_accounts.rb
ec: As soon, as there is no user side ability to create a service account, we allow full customization for the server logic.
Maybe its a good idea to add both ways, allow to pass UID, but if its nil - we can generate internally.
Works for you?

app/api/v2/management/service_accounts.rb
ec: If you need it, we can easily add it. Any more ideas for filtering?

dpatsora: Let's add filtering by provider_email and provider_uid. It will be enough. Thanks in advance)

app/api/v2/management/service_accounts.rb
dpatsora: I got your point. Great! 
I like this idea)

I think for internal generating we can define some prefix, that will differ with User one( E.g. "SA" instead of "ID")

I'm scared that, maybe user only signed up, without login. In this case, we don't have such User on Applogic. And we can pass you an uid, that's already exists on Barong

app/models/document.rb
dpatsora: I've thought about it)

app/models/user.rb
dpatsora: Np)

app/api/v2/management/service_accounts.rb
ec: @dpatsora yeah, dont be so scared, my little. There is a check inside ServiceAccount model. We always check if UID and EMAIL are uniq in both tables  - Users and ServiceAccounts

dpatsora: I'm scared not about Barong)
I just don't want to send one more request from Applogic and do it in loop)

db/migrate/20200818135531_add_service_accounts_and_related_changes.rb
```diff
@@ -0,0 +1,19 @@
...
+      t.bigint    :provider_id,         null: false, unsigned: true
+      t.string    :email,               null: false
+      t.string    :role,                default: "service_account", null: false
+      t.integer   :level,               default: 0, null: false
+      t.string    :state,               default: "pending", null: false
+
+      t.timestamps
+    end
+
+    remove_column :apikeys, :user_id
```
dpatsora: We loose here relation between already existing User and API_KEY

config/seeds.yml
```diff
@@ -263,3 +263,21 @@ permissions:
...
+ # has standard user permissions to ACCOUNT, MARKET, RESOURCE modules and RESOURCE of Service Accounts
+ - { role: provider, verb: all, path: api/v2/peatio/account, action: accept }
+ - { role: provider, verb: all, path: api/v2/peatio/market, action: accept }
+ - { role: provider, verb: all, path: api/v2/barong/resource, action: accept }
+ # ranger access
+ - { role: provider, verb: get, path: api/v2/ranger, action: accept }
+ # applogic access
+ - { role: provider, verb: all, path: api/v2/applogic, action: accept }
+
+ # Service Account
```
dpatsora, line 277: I'm not sure how to specify paths we need for service accounts
@ec @mnaichuk can you help me here?

app/api/v2/management/service_accounts.rb
```diff
@@ -0,0 +1,92 @@
...
+          status 200
+        end
+
+        desc 'Get service_accounts as a paginated collection' do
+          @settings[:scope] = :read_service_accounts
+          success API::V2::Entities::ServiceAccounts
+        end
+        params do
+          use :pagination_filters
+          optional :provider_uid, type: String, allow_blank: false, desc: 'provider uid'
```
calj: provider naming is wrong, it has to be generic, we can call it owner.

app/models/service_account.rb
```diff
@@ -0,0 +1,92 @@
...
+  def email_n_uid_uniqueness
+    errors.add('email_or_uid', 'not_uniq') if User.find_by(email: email) || User.find_by(uid: uid)
+  end
+
+  private
+
+  def assign_email
+    return unless email.blank?
+
+    self.email = "service+#{user.service_accounts.count}+" + user.email
```
calj: name, domain = user.email.split('@')
"#{name}+#{self.uid}@#{domain}"

app/api/v2/management/service_accounts.rb
```diff
@@ -0,0 +1,92 @@
...
+        end
+
+        desc 'Create service_account' do
+          @settings[:scope] = :write_service_accounts
+          success API::V2::Entities::ServiceAccounts
+        end
+        params do
+          optional :service_account_uid, type: String, allow_blank: false, desc: 'service_account uid'
+          requires :provider_uid, type: String, allow_blank: false, desc: 'provider uid'
+          optional :service_account_email, type: String, allow_blank: false, desc: 'service_account email'
```
calj, line 56: we need to set the role here

app/models/service_account.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  def assign_uid
+    return unless uid.blank?
+
+    loop do
+      self.uid = random_uid
+      break unless User.where(uid: uid).any? || ServiceAccount.where(uid: uid).any?
```
calj: replace this using "if" instead of "unless"

calj: and the same logic should be in user, so let's move it to a helper

app/api/v2/resource/providers.rb
```diff
@@ -0,0 +1,187 @@
...
+          end
+          error!({ errors: ['resource.providers.missing_totp'] }, 422) unless params[:totp_code].present?
+
+          return if TOTPService.validate?(current_user.uid, params[:totp_code])
+
+          error!({ errors: ['resource.providers.invalid_totp'] }, 422)
+        end
+      end
+
+      resource :providers do
```
dpatsora: Which name we will use here?

app/models/service_account.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  validate :role_exists
+  validate :email_n_uid_uniqueness
+  validates :email,       email: true, presence: true, uniqueness: true
+  validates :uid,         presence: true, uniqueness: true
+
+  scope :active, -> { where(state: 'active') }
+
+  after_update :disable_api_keys
+  before_create :assign_state
```
dpatsora, line 32: Maybe we need to disable all service accounts if user account has been disabled?

PR #1138, opened 2020-08-26
Add security.md

ec:


mod:
The security hardening should be in the form of a checklist
Also is there a way to set rate limit? or maybe other security params we recommend?

PR #1149, opened 2020-09-30
Fix: Mailer reinitialize on lost db connection instead of raise error

vpetrusenko:


app/services/event_mailer.rb
```diff
@@ -115,7 +115,11 @@ def handle_message(delivery_info, _metadata, payload)
   rescue StandardError => e
     Rails.logger.error { e.inspect }
 
-    raise e if is_db_connection_error?(e)
+    reconnect if is_db_connection_error?(e)
```
calj: here you need to make sure the daemon crash.
it's the most safe way to make such we reconnect and reinitialize everything.

vpetrusenko: sorry if didn't get it correctly, i update the commit and now i check the exception message to be equal to 'MySQL server has gone away'

PR #1150, opened 2020-09-30
Feature: Add encryption on profile, phone and document models

chumaknadya:


app/models/phone.rb
```diff
@@ -33,6 +38,15 @@ def valid?(unsafe_phone)
     def international(unsafe_phone)
       parse(unsafe_phone).international(false)
     end
+
+    def find_by_number(number, attrs={})
+      attrs.merge!(number_index: Zlib::crc32(number + Barong::App.config.crc32_salt))
```
calj: let's make a helper for generating the hash

app/helpers/crc32.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class Crc32
```
calj: Let's rename to SaltedCrc32

PR #1202, opened 2020-12-21
Feature: Mailer improvements

calj:
Fix event-api loading problem in development environment

- Remove Spring
- Add a development guide to use mailer locally
- Retry message in case of failure


app/services/event_mailer.rb
```diff
@@ -27,15 +27,34 @@ def listen
     end
 
     @bunny_channel = @bunny_session.channel
+    # Define fanout exchanges which will broadcast
+    # all the messages they receives to all the queues they know
+    retry_exchange = @bunny_channel.fanout('retry.exchange')
```
calj: 1. let's delete the old queue named postmaster something
2. prefix those exchanges with something like barong.event.mailer
3. rename main mailer queue to barong.event.mailer.main
4. rename retry mailer queue to barong.event.mailer.retry

config/initializers/barong.rb
```diff
@@ -8,27 +8,29 @@
...
-    raise Barong::KeyStore::Fatal
-  end
-rescue Barong::KeyStore::Fatal
-  Rails.logger.fatal('Private key is invalid')
-  raise 'FATAL: Private key is invalid'
+private_key_path = ENV['JWT_PRIVATE_KEY_PATH']
+
+if !private_key_path.nil?
+  pkey = Barong::KeyStore.open!(private_key_path)
+  Rails.logger.error('Loading private key from: ' + private_key_path)
```
dapi, line 15: @calj HI!

It this real error? Could be Rails.logger.info here?

PR #1217, opened 2021-01-06
Feature: Add IP country for admin user activities api

jengjeng:
Add IP country for admin user activities API.
This feature related to display IP country on the user activities page in Tower.


chumaknadya (approved):

app/api/v2/admin/entities/activity_with_user.rb
```diff
@@ -4,6 +4,9 @@ module API::V2::Admin
   module Entities
     class ActivityWithUser < API::V2::Entities::Base
       expose :user_ip, documentation: { type: 'String' }
+      expose :user_ip_country, documentation: { type: 'String' } do |activity|
```
chumaknadya, line 7: Let's add new field for the activity table, where we will define user_ip_country
Do not forget to add it here https://github.com/openware/barong/blob/master/lib/barong/activity_logger.rb#L37 too then

After adding new field, you don't need to do search by IP on every exposion

chumaknadya (changes requested):

josadcha:
hello @jengjeng 
Would be nice to have this PR finished, because we just found out that Tower doesn't display info about country at all. Just a kind reminder for you.

PR #1219, opened 2021-01-13
Fix: Prevent label duplication

chumaknadya:


db/migrate/20210113091207_add_unique_index_on_labels.rb
```diff
@@ -0,0 +1,6 @@
+class AddUniqueIndexOnLabels < ActiveRecord::Migration[5.2]
+  def change
+    remove_index :labels, [:user_id, :key, :scope]
+    add_index :labels, [:user_id, :key, :scope], unique: true
```
calj, line 12: You will have a problem here if there are already duplicated entries in the database, you can try to write a spec for this migration.

db/migrate/20210113091207_add_unique_index_on_labels.rb
```diff
@@ -0,0 +1,10 @@
+class AddUniqueIndexOnLabels < ActiveRecord::Migration[5.2]
+  def change
+    remove_index :labels, [:user_id, :key, :scope]
+
+    # Delete dublicated values
+    ActiveRecord::Base.connection.execute("DELETE FROM labels WHERE ID NOT IN (SELECT MAX(ID) from (SELECT * FROM labels) AS l GROUP BY l.user_id, l.scope, l.key )")
```
calj: It might be a bit brutal for the server if there is a lot of labels.

PR #1224, opened 2021-01-19
Add nickname into signup & users

xenirio:


app/models/user.rb
```diff
@@ -21,6 +21,7 @@ class User < ApplicationRecord
   validate :referral_exists
   validates :data, data_is_json: true
   validates :email,       email: true, presence: true, uniqueness: true
+  validates :nickname,    length: { minimum: 4, maximum: 12 }, format: { with: /\A[a-z0-9]+\z/ }, uniqueness: true
```
calj: we should probably force to lowercase somewhere

calj:
- [x] Add the nickname to the JWT , here: https://github.com/openware/barong/blob/5350260072ba86228882c10c30ad33771b036c04/app/models/user.rb#L163

db/migrate/20210119082215_add_nickname_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddNicknameToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :nickname, :string
```
calj: we need a unique index on this field.

app/models/user.rb
xenirio: OK. I will allow with the `format: { with: /\A[a-zA-Z0-9]+\z/ }` and force lowercase by:

```ruby
before_save :downcase_fields

def downcase_fields
  self.nickname.downcase!
end
```

app/api/v2/identity/users.rb
```diff
@@ -51,6 +51,10 @@ def parse_refid!
           { code: 422, message: 'Validation errors' }
         ]
         params do
+          requires :nickname,
```
chumaknadya: should be optional

app/api/v2/entities/user_with_profile.rb
```diff
@@ -4,6 +4,7 @@ module API::V2
   module Entities
     class UserWithProfile < API::V2::Entities::Base
       expose :email, documentation: { type: 'String' }
+      expose :nickname, documentation: { type: 'String' }
```
chumaknadya: please put description

app/models/user.rb
```diff
@@ -21,6 +21,7 @@ class User < ApplicationRecord
   validate :referral_exists
   validates :data, data_is_json: true
   validates :email,       email: true, presence: true, uniqueness: true
+  validates :nickname,    length: { minimum: 4, maximum: 12 }, format: { with: /\A[a-zA-Z0-9]+\z/ }, uniqueness: true
```
chumaknadya: validate if present
we dont need validation for nil nickname

app/models/user.rb
```diff
@@ -30,9 +31,14 @@ class User < ApplicationRecord
                                            { key: 'document', value: ['pending', 'replaced'], scope: 'private' }) }
 
   before_validation :assign_uid
+  before_save :downcase_fields
   after_update :disable_api_keys
   after_update :disable_service_accounts
 
+  def downcase_fields
```
chumaknadya: rename function as you downcase only nickname


And what will be here if you try to downcase nil

chumaknadya (changes requested):

db/migrate/20210119082215_add_nickname_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddNicknameToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :nickname, :string, after: :email
+    add_index :users, :nickname, unique: true
```
chumaknadya: How unique field will work if there is nil value?

chumaknadya (changes requested):

app/api/v2/entities/service_accounts.rb
```diff
@@ -4,6 +4,7 @@ module API::V2
   module Entities
     class ServiceAccounts < API::V2::Entities::Base
       expose :email, documentation: { type: 'String' }
+      expose :nickname, documentation: { type: 'String' }
```
chumaknadya: why you should have nickname on service account?

app/api/v2/entities/user.rb
```diff
@@ -6,6 +6,7 @@ module Entities
       # Basic user info
       class User < API::V2::Entities::Base
         expose :email, documentation: { type: 'String' }
+        expose :nickname, documentation: { type: 'String' }
```
chumaknadya: this field should be added also on admin entities and management API entities

chumaknadya (changes requested):

app/api/v2/entities/user.rb
xenirio: I think the nickname filed already exposed in admin/mannagement API entities because we use `expose :user, using: API::V2::Entities::User`

chumaknadya: true

app/api/v2/entities/user_with_profile.rb
```diff
@@ -4,6 +4,7 @@ module API::V2
   module Entities
     class UserWithProfile < API::V2::Entities::Base
       expose :email, documentation: { type: 'String' }
+      expose :nickname, documentation: { type: 'String', desc: 'user nickname' }
```
chumaknadya: Can you add description everywhere in entities with first Upercase letter, for example with entity on kyc etc

db/migrate/20210119082215_add_nickname_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddNicknameToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :nickname, :string, null: true, after: :email
```
chumaknadya: you should manage uniqueness through db too
So it should be uniq if value is not nil, and we shouldn't check uniqueness if value nil

chumaknadya (changes requested):

chumaknadya:
Also can you write few sentences on our documentation that now we have ability to use nickname ?

db/migrate/20210119082215_add_nickname_to_users.rb
```diff
@@ -0,0 +1,6 @@
...
+    add_index :users, :nickname, unique: true, where: 'nickname IS NOT NULL'
```
chumaknadya: i think we should leave after: :email 
also prev it was :string type, should we specify type here?

chumaknadya (changes requested):

db/migrate/20210119082215_add_nickname_to_users.rb
xenirio: Can you explain for about :string type? Is the add_cloumn I do wrong format?

chumaknadya: ah, sorry, i didnt check line before

chumaknadya (approved):

xenirio:
I fixed the bug of nickname not validate uniqueness of case-sensitive nickname value. Please check. Thank you.

db/migrate/20210119082215_add_nickname_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddNicknameToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :nickname, :string, null: true
+    add_index :users, :nickname, unique: true, where: 'nickname IS NOT NULL'
```
chumaknadya: please add tests for this case

1) try to insert 2 user with same nickname
2) try to insert 2 user without nickname (check if there is no validation in uniq for such case)

app/models/user.rb
```diff
@@ -29,10 +30,15 @@ class User < ApplicationRecord
   scope :with_pending_or_replaced_docs, -> { self.joins(:labels).where(labels:
                                            { key: 'document', value: ['pending', 'replaced'], scope: 'private' }) }
 
-  before_validation :assign_uid
+  before_validation :assign_uid, :downcase_nickname
```
chumaknadya: why you need both before save amd before validation
Maybe before validation should just work

chumaknadya (changes requested):

app/models/user.rb
xenirio: Ah, OK. Let me do this.

db/migrate/20210119082215_add_username_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddUsernameToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :username, :string, null: true
+    add_index :users, :username, unique: true, where: 'username IS NOT NULL'
```
calj, line 4: Is this `where` doing anything?
Can you show me the `show create table` ?

xenirio: The where clause will check to validate uniqueness for not-null username. I tested it works. ref: https://stackoverflow.com/a/48147333/5087746.

here is `show create table`
```sql
CREATE TABLE `users` (
  `id` bigint NOT NULL AUTO_INCREMENT,
  `uid` varchar(255) NOT NULL,
  `email` varchar(255) NOT NULL,
  `password_digest` varchar(255) NOT NULL,
  `role` varchar(255) NOT NULL DEFAULT 'member',
  `data` text,
  `level` int NOT NULL DEFAULT '0',
  `otp` tinyint(1) DEFAULT '0',
  `state` varchar(255) NOT NULL DEFAULT 'pending',
  `referral_id` bigint DEFAULT NULL,
  `created_at` datetime NOT NULL,
  `updated_at` datetime NOT NULL,
  `username` varchar(255) DEFAULT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `index_users_on_uid` (`uid`),
  UNIQUE KEY `index_users_on_email` (`email`),
  UNIQUE KEY `index_users_on_username` (`username`)
) ENGINE=InnoDB AUTO_INCREMENT=3 DEFAULT CHARSET=utf8
```

PR #1229, opened 2021-01-20
Admin API for maintenance page

mnaichuk:
TODO:
- [x] add api entities
- [x] add GET and PUT to the admin controller
- [ ] add specs for model, workers and API
- [ ] Add Job ability to the default values in the charts

### DIFF
--- a/app/api/v2/admin/jobs.rb
+++ b/app/api/v2/admin/jobs.rb
@@ -113,8 +113,13 @@ module API
                      allow_blank: false,
                      desc: 'Job id'
             requires :state,
+                     type: String,
                      allow_blank: false,
                      values: { value: -> { Job::STATES }, message: 'admin.job.invalid_state' }
+            optional :description,
+                     type: String,
+                     allow_blank: false,
+                     desc: 'job description'
             optional :start_at,
                      type: DateTime,
                      desc: 'date and time to run start job'
@@ -122,7 +127,7 @@ module API
                      type: DateTime,
                      desc: 'date and time to run finish job'
           end
-          put do
+          put '/:id' do
             admin_authorize! :update, Job
 
             target_job = Job.find_by(id: params[:id])
--- a/app/api/v2/admin/jobs.rb
+++ b/app/api/v2/admin/jobs.rb
@@ -38,7 +38,7 @@ module API
             admin_authorize! :create, Job
             admin_authorize! :create, Jobbing
             
-            # Find or create maintenace restriction
+            # Find or create maintenance restriction
             maintenance = Restriction.find_or_create_by(category: params[:type], scope: 'all', value: 'all', state: 'disabled')
             error!({ errors: ['admin.job.cant_find_or_create_maintenance_restriction'] }, 422) unless maintenance.id.present?
             
@@ -135,7 +135,7 @@ module API
 
             # clear cached restrictions, so they will be freshly refetched on the next call to /auth
             Rails.cache.delete('restrictions')
-            
+
             status 200
           end
         end
--- a/app/api/v2/admin/jobs.rb
+++ b/app/api/v2/admin/jobs.rb
@@ -20,43 +20,57 @@ module API
                      desc: 'job description'
             requires :type,
                      type: String,
-                     values: { value: -> { ['maintenance'] }, message: 'admin.job.invalid_type'},
+                     values: { value: -> { Job::TYPES }, message: 'admin.job.invalid_type'},
                      desc: 'job type'
             requires :start_at,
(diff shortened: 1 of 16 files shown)


db/schema.rb
```diff
@@ -77,6 +77,19 @@
     t.index ["user_id"], name: "index_documents_on_user_id"
   end
 
+  create_table "jobs", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
+    t.string "reference_type", null: false
+    t.bigint "reference_id", null: false
+    t.integer "type", null: false
+    t.text "description", null: false
+    t.integer "state", null: false
```
kengio, line 85: I think It should be string. Right?

db/migrate/20210120100145_create_jobs.rb
```diff
@@ -0,0 +1,15 @@
+class CreateJobs < ActiveRecord::Migration[5.2]
+  def change
+    create_table :jobs do |t|
```
mnaichuk, line 3: we discussed with Camille that we will need to have a separate relation table, smth like:
job_id
reference_type
reference_id

This will help us to solve:
1. When one job has two restrictions (for maintenance and for whitelist during the maintenance)
2. Keep flexibility here to have the ability to create Job not only for Restrictions

kengio: I create a new intermediate table named `jobbings` as picture below.
Barong Jobs

I can create Job, Restriction and relationships like this
```ruby

# Get or create maintenance restriction
maintenance = Restriction.find_or_create_by(category: 'maintenance', scope: 'all', value: 'all', state: 'disabled')

# Get or create whitelist restriction for admin
whitelist = Restriction.find_or_create_by(category: 'whitelist', scope: 'ip', value: '164.116.47.255', state: 'disabled')

# Create new maintenance job
job = Job.create!(type: 'maintenance', description: 'monthly maintenance', state: 'pending', start_at: "2020-01-31T00:00:00", finish_at: '2020-01-31T03:00:00')

# Create relationship from job to maintenance and whitelist restrictions
Jobbing.create!(job: job, reference: maintenance)
Jobbing.create!(job: job, reference: whitelist)

# Get all restrictions by job
restrictions = job.restrictions # or job.jobbings

# Get all jobs by restriction
jobs = maintenance.jobs # or maintenance.jobbings

# Get all pending jobs by restriction
pending_jobs = jobs.pending

# Update all pending jobs by restriction to active
pending_jobs.map(&:active!)

# Update all restrictions by job to enabled
restrictions.update_all(state: 'enabled')
````

If other models need to create new Job it can use polymorphic relationship feature in Rails.

@mnaichuk 
My questions are:
1. Do we need to create new restriction for each job or reuse existing restriction? If we reuse existing restriction what happen if some jobs want to enabled/disabled the restriction and conflict with other jobs. 
2. Do we still need `jobs.type` field if we alway create maintenance and whitelist restrictions together? Or you mean `maintenance` is only type we support now.

File: db/migrate/20210120100145_create_jobs.rb [hunk 3]
```diff
@@ -0,0 +1,12 @@
...
+      t.integer    :type,        null: false # enum ( whitelist: 0, maintenance: 1, blacklist: 2, blocklogin: 3 )
```
kengio: @mnaichuk 

Do we allow only `maintenance` type?

app/api/v2/admin/jobs.rb
```diff
@@ -0,0 +1,83 @@
...
+          params do
+            requires :description,
+                     type: String,
+                     allow_blank: false,
+                     desc: 'job description'
+            requires :type,
+                     type: String,
+                     values: { value: -> { Job::TYPES }, message: 'admin.job.invalid_type'},
+                     desc: 'job type'
+            requires :start_at,
```
mnaichuk, line 25: can be optional

kengio: How to schedule the worker If `start_at` is nil?

Do you mean to have logic like this? And How about `finish_at`?

```ruby
if start_at.nil?
  Jobs::JobStartWorker.perform_async(to_sgid(for: 'job_start'))
else
  Jobs::JobStartWorker.perform_at(start_at, to_sgid(for: 'job_start'))
end
```

mnaichuk: oh, sorry, finish_at can optional, you can schedule maintenance without end date and whitelist your IP to disable maintenance from Tower UI

kengio: Changed.

PR #1234, opened 2021-01-26
Feature: Add username to user's model

xenirio:


config/backend.yml
```diff
@@ -3,7 +3,7 @@ version: '3'
 services:
 
   db:
-    image: mysql:5.7
+    image: mysql:8.0
```
chumaknadya: I don't think we need to change mysql image in this PR

chumaknadya:
Also be sure you don't have private and public keys in this PR

config/backend.yml
ghost: Okay
I reverted this change

db/migrate/20210316083841_add_username_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddUsernameToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :username, :string, null: true
```
chumaknadya: username should be after uid field

``add_column :users, :username, :string, null: true, after: :uid``

chumaknadya (changes requested):

chumaknadya (approved):

PR #1235, opened 2021-01-26
Feature: add order_by parameter to get labels endpoint

vpetrusenko:


app/api/v2/resource/labels.rb
```diff
@@ -12,7 +12,8 @@ class Labels < Grape::API
                  { code: 401, message: 'Invalid bearer token' }
                ]
           get do
```
chumaknadya, line 21: add params order_by here 
Also, please add spec

chumaknadya (changes requested):

app/api/v2/resource/labels.rb
vpetrusenko: @chumaknadya updated

chumaknadya:
Are you sure it should be on master branch instead of 2-6-stable ?

app/api/v2/resource/labels.rb
```diff
@@ -11,8 +11,16 @@ class Labels < Grape::API
...
+          params do
+            requires :order_by,
+                     type: String,
+                     allow_blank: true,
+                     desc: 'Parameter for ordering.'
+          end
           get do
-            present current_user.labels, with: Entities::Label
+            order = params[:order_by]
+            labels = order ? current_user.labels.order("#{order} DESC") : current_user.labels
```
chumaknadya: should be smth like
```
          optional :order_by,
                   type: String,
                   values: { value: %w(asc desc) },
                   default: 'desc',
                   desc: 'If set, returned labels  sorted in specific order, default to "desc".'
  ```     
  ```
  current_user.labels.order(created_at: params[:order_by])
  ```

File: spec/api/v2/resource/labels_spec.rb [hunk 3]
```diff
@@ -18,17 +18,24 @@
       value: ::Faker::Internet.slug(glue: '-')
     }
   end
+  let(:order_by_param) do
+    {
+      order_by: 'created_at'
```
chumaknadya: in this case you dont need param here

chumaknadya (changes requested):

PR #1237, opened 2021-01-28
Fix: default ordering for labels to asc

vpetrusenko:


app/api/v2/resource/labels.rb
```diff
@@ -15,7 +15,7 @@ class Labels < Grape::API
             optional :ordering,
                      type: String,
                      values: { value: %w(asc desc) },
-                     default: 'desc',
+                     default: 'asc',
                      desc: 'If set, returned labels sorted in specific order, default to "desc".'
```
chumaknadya: default to 'asc'

chumaknadya (changes requested):

app/api/v2/resource/labels.rb
chumaknadya: also can you check rspec ?

chumaknadya (approved):

PR #1246, opened 2021-02-16
Feature: Add ability to register first user with superadmin role

chumaknadya:


config/initializers/barong.rb
```diff
@@ -57,6 +57,7 @@
   config.set(:profile_double_verification, 'false', type: :bool)
   config.set(:crc32_salt, '')
   config.set(:api_data_masking_enabled, 'true', type: :bool)
+  config.set(:first_superadmin_registration, 'false', type: :bool)
```
calj: let's rename it :first_registration_superadmin and default true

app/api/v2/identity/users.rb
```diff
@@ -78,12 +78,18 @@ def parse_refid!
...
 
           user = User.new(user_params)
-
           code_error!(user.errors.details, 422) unless user.save
 
           activity_record(user: user.id, action: 'signup', result: 'succeed', topic: 'account')
 
-          publish_confirmation(user, Barong::App.config.domain)
+          # Creates superadmin user in first platform registration
+          if Barong::App.config.first_superadmin_registration && user.id == 1
```
calj: It's better you check the number of entries in database:
```
User.count == 1
```
The id would not be one if we truncate the table.

docs/configuration.md
```diff
@@ -18,6 +18,7 @@
 | `barong_jwt_expire_time` | '3600' | integer representation of seconds  | general purpose tokens (reset password, confirm email) expiration time |
 | `crc32_salt` | - | any string value | salt for crc32 algorithm which used to searching in encrypted fields |
 | `api_data_masking_enabled` | true | `true`, `false` | while turned on (`true`) user API will be with ecnrypted user data |
+|`first_superadmin_registration`| false | `false` `true` | while turned on (`true`) first registered user on a platform will be superadmin without any email confirmation |
```
calj: "when turned on" sounds more correct than "while turned on"
You can change the line before too

PR #1251, opened 2021-02-25
Feature: Add management API endpoints to create/update APIkeys

chumaknadya:


app/api/v2/management/api_keys.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+module API::V2
+  module Management
+    class APIKeys < Grape::API
```
calj, line 5: if this is specific to service account we should probably prefix the endpoint with "sa" or something

calj: or we rename uid to sa_uid so we could use this endpoint for managing either user api_keys or service account.

app/api/v2/management/api_keys.rb
```diff
@@ -0,0 +1,98 @@
...
+                   allow_blank: false,
+                   desc: 'User UID or Service Account UID'
+          optional :scopes,
+                   type: String,
+                   allow_blank: false,
+                   desc: 'Comma separated scopes'
+        end
+        post do
+          if params[:uid].start_with?(Barong::App.config.uid_prefix)
+            key_holder = User.find_by(uid: params[:uid])
```
calj, line 29: let's introduce a configuration entry to enable this feature, and we split it in 2:
mgn_api_keys_users
mgn_api_keys_sa
we default them to false so by default this is disabled.

PR #1261, opened 2021-03-15
Feature: Add multiplatform support

chumaknadya:
- [ ] registration
- [ ] login
- [x] jwt (add platform id and hostname)

### DIFF
--- a/app/models/platform.rb
+++ b/app/models/platform.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+class Platform < ApplicationRecord
+  def self.platform_by_host(host)
+    return if Platform.count == 0
+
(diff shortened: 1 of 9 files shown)


app/models/platform.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class Platform < ApplicationRecord
+	def self.platform_by_host(host)
+		return if Platform.count == 0
```
calj: here we would default to one

chumaknadya: I created platfrom_id as an optional value for user table
So, if there is no platform in db we will assign nil by default
Cause MySQL cant link to platform table with non-existent ID

PR #1279, opened 2021-04-07
Feature: Add attachments

mnaichuk:


app/api/v2/management/entities/attachment.rb
```diff
@@ -0,0 +1,23 @@
...
+module API::V2
+  module Management
+    module Entities
+      class Attachment < API::V2::Entities::Base
+        expose :id,
+               documentation: {
+                 type: 'Integer',
+                 desc: 'Activity ID'
+               }
+        expose :uid,
```
chumaknadya: maybe lets expose :user_uid instead of :uid
To not confuse users

app/api/v2/management/entities/attachment.rb
```diff
@@ -0,0 +1,23 @@
...
+               if: ->(attachment, _options) { attachment.user },
+               documentation: {
+                 type: 'Strinf',
```
chumaknadya: String

chumaknadya (changes requested):

PR #1289, opened 2021-04-14
script for creating labels

crm-skyach:


calj:
You implement the specific business case here, this task should be generic.
It should take the label to apply in argument, then apply it for each uid from the csv.

crm-skyach:
Hi @calj I have made the changes according to your comment, please check if this is what you are expecting.

chumaknadya (approved):

PR #1293, opened 2021-04-19
Feature: Separate request domain and protocol

xenirio:


app/workers/kyc/kycaid/document_worker.rb
```diff
@@ -76,10 +76,11 @@ def selfie_image_params(selfie_image)
       end
 
       def verification_params
+        schema = Barong::App.config.tls_enabled ? 'https' : 'http'
         {
           applicant_id: @applicant_id,
           types: ['DOCUMENT', 'FACIAL'],
-          callback_url: "#{Barong::App.config.domain}/api/v2/barong/public/kyc"
+          callback_url: "#{schema}://#{Barong::App.config.domain}/api/v2/barong/public/kyc"
```
chumaknadya: We come with a little improvement
It seems it will be like a Breaking Change for our users
So maybe lets do like this:
Check if #{Barong::App.config.domain} has https/http schema, you should get only domain from this URI, if there is no uri put as it is

app/workers/kyc/kycaid/document_worker.rb
```diff
@@ -76,10 +76,11 @@ def selfie_image_params(selfie_image)
       end
 
       def verification_params
+        schema = Barong::App.config.tls_enabled ? 'https' : 'http'
```
chumaknadya: We also use Barong domain in our email templates
There we should do this validation too

chumaknadya (changes requested):

config/initializers/barong_domain.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Barong
+  class Config
+    class << self
+      def domain
+        app_domain = Barong::App.config.domain
```
dpatsora: Would be cool to have `Barong::App.config.url`

config/initializers/barong_domain.rb
```diff
@@ -0,0 +1,16 @@
...
+        unless app_domain.match? %r{^http(s?)://}
```
dpatsora: if

config/initializers/barong_domain.rb
```diff
@@ -0,0 +1,16 @@
...
+          schema = Barong::App.config.tls_enabled ? 'https' : 'http'
+          app_domain = "#{schema}://#{Barong::App.config.domain}"
+        end
+        app_domain
```
dpatsora: rename app_domain to app_url

config/initializers/barong_domain.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Barong
+  class Config
+    class << self
+      def domain
```
dpatsora: url

config/initializers/barong_domain.rb
xenirio: I can't override the `Barong::App.config.url` as a method. So, I will use `Barong::App.url` instead.

app/api/v2/identity/users.rb
```diff
@@ -163,7 +163,8 @@ def parse_refid!
             EventAPI.notify('system.user.email.confirmed',
                             record: {
                               user: current_user.as_json_for_event_api,
-                              domain: Barong::App.config.domain
```
chumaknadya, line 166: we don't need  domain anymore

app/api/v2/identity/users.rb
```diff
@@ -271,7 +273,8 @@ def parse_refid!
             EventAPI.notify('system.user.password.reset',
                             record: {
                               user: current_user.as_json_for_event_api,
```
chumaknadya, line 273: we don't need domain anymore

app/api/v2/identity/utils.rb
```diff
@@ -99,13 +99,14 @@ def token_uniq?(jti)
         Rails.cache.write(jti, 'utilized', expires_in: Barong::App.config.jwt_expire_time.seconds)
       end
 
-      def publish_confirmation(user, domain)
```
chumaknadya, line 97: domain from here we should delete

app/api/v2/resource/users.rb
```diff
@@ -148,7 +148,8 @@ def verify_otp!
           EventAPI.notify('system.user.password.change',
                           record: {
                             user: current_user.as_json_for_event_api,
-                            domain: Barong::App.config.domain
```
chumaknadya, line 151: the same

chumaknadya (changes requested):

PR #1304, opened 2021-04-21
3-0-stable Feature: Separate request domain and protocol

xenirio:


app/api/v2/resource/users.rb
```diff
@@ -148,7 +148,7 @@ def verify_otp!
           EventAPI.notify('system.user.password.change',
                           record: {
                             user: current_user.as_json_for_event_api,
-                            domain: Barong::App.config.domain
+                            url: Barong::App.url
```
dpatsora, line 151: I would keep both fields to not introduce breaking changes

PR #1315, opened 2021-05-21
Feature : Introduce organization feature

xenirio:


app/api/v2/commercial/organizations.rb
```diff
@@ -0,0 +1,174 @@
...
+                     desc: 'organization postcode'
+          end
+          post '/update' do
+            admin_organization_authorize!
+
+            target_organization = Organization.find_by_oid(params[:oid])
+
+            # Ruby Hash returns array on keys and values
+            update_param_key = params.except(:oid).keys.first
+            update_param_value = params.except(:oid).values.first
```
dnfd: did not get it

app/api/v2/identity/utils.rb
```diff
@@ -27,6 +27,26 @@ def open_session(user)
         csrf_token
       end
 
+      def open_session_switch(user, switch)
+        csrf_token = SecureRandom.hex(10)
+        expire_time = Time.now.to_i + Barong::App.config.session_expire_time
+        session.merge!(
+          "uid": user.uid,
```
dnfd: should it be org_id instead?

spec/api/v2/commercial/organizations_spec.rb
```diff
@@ -0,0 +1,244 @@
...
+        post '/api/v2/commercial/organizations',
+             params: { name: 'Company Test', group: 'vip-1' },
+             headers: auth_header
+
+        expect(response.status).to eq(404)
+      end
+
+      it 'create organization for barong organization admin' do
+        # Assign user as barong organization admin
+        create(:membership, id: 1, user_id: 1, organization_id: 0)
```
dnfd: what is zero organization?

app/models/membership.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class Membership < ApplicationRecord
+  belongs_to :organization, optional: true
```
dnfd: sounds like it should be required

spec/api/v2/commercial/organizations_spec.rb
```diff
@@ -0,0 +1,244 @@
...
+      end
+
+      it 'return organization details for barong organization admin' do
+        # Assign user as barong organization admin
+        create(:membership, id: 1, user_id: 1, organization_id: 0)
+
+        do_request
+
+        expect(response.status).to eq(200)
+      end
```
dnfd: status checked, response is not

spec/api/v2/commercial/organizations_spec.rb
```diff
@@ -0,0 +1,244 @@
...
+      it 'update organization field' do
+        # Assign user as barong organization admin
+        create(:membership, id: 1, user_id: 1, organization_id: 0)
+
+        params[:name] = 'Company Test'
+        do_request
+
+        expect(response.status).to eq(200)
+      end
+
```
dnfd: database state is not checked, cannot ensure request did anything

dnfd: testcase with several params changed is missing, and from the code it looks like A\pi can only update one key at a time

spec/api/v2/commercial/organizations_spec.rb
dnfd: ah, so is it ephemerial platform org?

spec/api/v2/commercial/organizations_spec.rb
```diff
@@ -0,0 +1,244 @@
...
+      it 'update organization group' do
+        # Assign user as barong organization admin
+        create(:membership, id: 1, user_id: 1, organization_id: 0)
+
+        params[:group] = 'vip-1'
```
dnfd: what if unexisting fee group assigned

spec/api/v2/identity/organization_session_spec.rb
```diff
@@ -0,0 +1,260 @@
...
+          # Assign user as organization admin
+          create(:membership, id: 1, user_id: 1, organization_id: 0)
+        end
+
+        it 'can switch to organization admin of Company A' do
+          params[:oid] = 'OID001'
+          switch_session
+
+          expect_status_to_eq 200
+          expect(session.instance_variable_get(:@delegate)[:aid]).to eq(params[:oid])
```
dnfd: should not it be uid?

app/api/v2/resource/utils.rb
```diff
@@ -11,6 +11,22 @@ def current_user
...
+        else
+          raise(Peatio::Auth::Error, 'Middleware Error')
+        end
+      end
+
+      def current_account
+        if env[:current_payload].has_key?(:aid)
+          @_current_account ||= Organization.find_by!(oid: env[:current_payload][:aid])
+        else
+          raise(Peatio::Auth::Error, 'Middleware Error')
```
dnfd: peatio?

app/api/v2/commercial/accounts.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
```
calj: commercial is not correct, I think we would be in organizations scope.

app/api/v2/commercial/accounts.rb
```diff
@@ -0,0 +1,59 @@
...
+            # Check account in the organization that user belong to
+            members = Membership.joins('LEFT JOIN organizations ON organizations.id = memberships.organization_id')
+                                .where(user_id: current_user.id)
+                                .select('memberships.*,organizations.name, organizations.organization_id')
+                                .pluck(:organization_id, :'organizations.name', :'organizations.organization_id')
+                                .map { |id, name, pid| { id: id, name: name, pid: pid } }
+            error!({ errors: ['identity.member.not_found'] }, 404) if members.nil? || members.length.zero?
+
+            # Check user is barong organization admin or not
+            if members.first[:id].zero?
```
calj: this is very suspicious to me, what is the logic?

app/api/v2/identity/sessions.rb
```diff
@@ -136,6 +136,87 @@ def get_user(email)
...
+                     desc: 'Organization OID'
+          end
+          post do
+            user = User.find_by(uid: session[:uid])
+            error!({ errors: ['identity.session.not_found'] }, 404) unless user
+
+            aid = params[:oid]
+            unless aid.nil?
+              # Check account in the organization that user belong to
+              members = Membership.joins('LEFT JOIN organizations ON organizations.id = memberships.organization_id')
```
calj: this logic deserves a specific helper with unit tests.

app/api/v2/identity/sessions.rb
```diff
@@ -136,6 +136,87 @@ def get_user(email)
...
+                                  .where(user_id: user.id)
+                                  .select('memberships.*,organizations.name, organizations.organization_id')
+                                  .pluck(:organization_id, :'organizations.name', :'organizations.organization_id')
+                                  .map { |id, name, pid| { id: id, name: name, pid: pid } }
+              error!({ errors: ['identity.member.not_found'] }, 404) if members.nil? || members.length.zero?
+
+              # Check user is barong organization admin or not
+              if members.first[:id].zero?
```
calj: There is not only one user with the ability to switch to any organization, it's based on user role and abilities associated to it.

db/schema.rb
```diff
@@ -98,6 +98,34 @@
...
+    t.string "role", default: "member", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["organization_id"], name: "index_memberships_on_organization_id"
+    t.index ["user_id"], name: "index_memberships_on_user_id"
+  end
+
+  create_table "organizations", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
+    t.string "oid", null: false
+    t.bigint "organization_id"
```
calj: please rename to **parent_id**

app/api/v2/identity/sessions.rb
```diff
@@ -136,6 +136,87 @@ def get_user(email)
...
+                # User is barong organization admin
+                oids = Organization.all.pluck(:id)
+              else
+                # User is organizationn admin/account
+                oids = Organization.where(id: members.pluck(:id)).pluck(:id)
```
calj: To be refactored, you can get the current user membership and check the role.

calj:
We need to introduce new roles into abilities.yaml:
- org-admin
- org-member
- org-accountant

Then we need to create new abilities for platform operators to manage the organizations:

```yaml

admin_permissions:
  superadmin:
    manage:
      - Organization
      - AdminSwitchSession

```

spec/api/v2/commercial/organizations_spec.rb
xenirio: `organization_id: 0` is for admin who can access to all organizations

app/models/membership.rb
xenirio: This is for case of user is admin of all organizations. It's `organization_id: 0`

spec/api/v2/commercial/organizations_spec.rb
xenirio: Correct, This endpoint is allow only 1 param to update b/c the UI don't have save btn to make save all params at a time.

spec/api/v2/commercial/organizations_spec.rb
xenirio: This is no verify here for group. Do we need to get all group from `peatio` and validate at the backend?

app/api/v2/resource/utils.rb
xenirio: Code removed.

app/api/v2/commercial/accounts.rb
xenirio: I set the admin who can access to all organization with `organization_id: 0`. So I check the level from here.

app/api/v2/identity/sessions.rb
xenirio: Right now. I determine the level of user by check the membership relation that I think it make sense to organize the user in the organization. I'm worry about If we determine user permission/access from `role` column in the membership table the organization access level will not correct.

app/api/v2/commercial/accounts.rb
xenirio: @calj Please check the following endpoint. Is this correct?

List

1. List accounts which user have access.
GET `/api/v2/barong/organization/accounts`

2. List all organizations 
GET `/api/v2/barong/organization/organizations`

3. Get organization info
GET `/api/v2/barong/organization?oid=...`

4. List organization users
GET `/api/v2/barong/organization/users?oid=...`

5. List organization subunits
GET `/api/v2/barong/organization/organizations?oid=...`

Create

1. Create organization
POST `/api/v2/barong/organization`

2. Create subunit in the organization
POST `/api/v2/barong/organization/account`

3. Add existing user into organization/subunit
POST `/api/v2/barong/organization/user`

Update

1. Update organization info
POST `/api/v2/barong/organization/update`

2. Update organization settings
POST `/api/v2/barong/organization/setting`

Delete

1. Delete subunit in the organization
DELETE `/api/v2/barong/organization/account`

2. Delete user in the organization
DELETE `/api/v2/barong/organization/user`

spec/api/v2/commercial/organizations_spec.rb
xenirio: I will do more check to ensure request was successfully

app/models/membership.rb
xenirio: OK. I got confirm that no need to do admin of all organization with `organization_id: 0`.
I will change to required.

app/api/v2/commercial/accounts.rb
xenirio: OK. I will change the logic of check admin of all organization to be check from user role instead.

app/api/v2/identity/sessions.rb
xenirio: > To be refactored, you can get the current user membership and check the role.

I got you. Right now I integrated the user role with abilities.

To verify admin with Organization ability

```ruby
admin_organization? :read, Organization
```

To restrict the access to Organization

```ruby
admin_organization_authorize! :read, Organization
```

spec/api/v2/commercial/organizations_spec.rb
xenirio: Updates, I've already verify admin of all organization by role & abilities. No need for extra code here.

app/api/v2/identity/sessions.rb
xenirio: For `unless aid.nil?` I do check that if :
- `params[:oid]` provided that mean user need to switch to org mode.
- `params[:oid]` not provide that mean user need to remove org mode and turn to be individual user.

calj:
When an admin use the ability AdminSwitchSession to switch to a **user** we set the same user role to his session.
When an admin use the ability AdminSwitchSession to switch to an **organization** we set the role from a static configuration:
```
BARONG_ADMIN_SWITCH_SESSION_ORG_ROLE=member
```

app/api/v2/identity/sessions.rb
xenirio: You're right. I've updated code based on user role with abilities.

app/api/v2/identity/utils.rb
xenirio: Correct. Fixed.

spec/api/v2/identity/organization_session_spec.rb
xenirio: Spec updated.

app/api/v2/commercial/entities/organization.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module API::V2::Commercial
+  module Entities
+    class Organization < API::V2::Entities::Organization
+      expose :email,
+             documentation: {
+               type: 'String',
+               desc: 'Organization Conntact Email'
```
calj: typo

app/api/v2/identity/sessions.rb
```diff
@@ -111,29 +117,111 @@ def get_user(email)
...
+            # Check switch session mode. Switch proceed only if params[:oid] provided
+            unless switch_oid.nil?
+              # Verify barong admin has AdminSwitchSession permission
+              if admin_organization? :read, AdminSwitchSession
+                # User is barong organization admin
+                oids = Organization.all.pluck(:id)
+              else
+                # User is organization admin/account
+                # Check account in the organization that user belong to
+                members = Membership.joins('LEFT JOIN organizations ON organizations.id = memberships.organization_id')
```
calj: I would like to review the generated request, can you post an example in reply?

app/api/v2/identity/sessions.rb
```diff
@@ -111,29 +117,111 @@ def get_user(email)
...
 
-            activity_record(user: user.id, action: 'login', result: 'succeed', topic: 'session')
-            csrf_token = open_session(user)
-            publish_session_create(user)
+            activity_record(user: user.id, action: (switch_oid.nil? ? 'switch::user' : 'switch::orgnization'),
+                            result: 'succeed', topic: 'session')
+            Barong::RedisSession.delete(user.uid, session.id)
+            session.destroy
+
+            if switch.nil?
```
calj, line 338: I don't understand what we are doing here, we should stop the execution before if something is wrong in params.

app/api/v2/identity/sessions.rb
```diff
@@ -111,29 +117,111 @@ def get_user(email)
...
+                     type: String,
+                     desc: 'User UID'
+          end
+          post do
+            user = User.find_by_uid(user_uid)
+            error!({ errors: ['identity.session.not_found'] }, 404) unless user
+
+            switch_oid = params[:oid]
+            # Check switch session mode. Switch proceed only if params[:oid] provided
+            unless switch_oid.nil?
```
calj: better to stop the execution before if param is nil

xenirio: params[:oid] is allow nil. If it's nil means you need to switch back (destroy) as individual user.

app/api/v2/identity/sessions.rb
xenirio: This is for checking the switch mode. You can use the same endpoint to be switch or destroy switch mode.
It's depend on the calculation of `switch` from above process.

app/api/v2/identity/sessions.rb
xenirio: The result of `members` could be: 


1. If you're `admin` of `Company A` and try to switch as `account` in Company A
```ruby
[{:id=>1, :name=>"Company A", :pid=>nil}]
```

2. If you're `member` of `Company A` and try to switch as `account` in Company A
```ruby
[{:id=>3, :name=>"Group A1", :pid=>1}]
```

app/api/v2/commercial/organizations.rb
xenirio: This logic I try to get the updated value and check with existing value to verify that data was changed before do save.

app/api/v2/commercial/accounts.rb
chumaknadya: List

List accounts which user have access.
GET /api/v2/barong/organization/accounts

List all organizations
GET /api/v2/barong/organization/all

Get organization info
GET /api/v2/barong/organization?oid=...

List organization users
GET /api/v2/barong/organization/users?oid=...

List organization subunits
GET /api/v2/barong/organization/account?oid=...

Create

Create organization
POST /api/v2/barong/organization

Create subunit in the organization
POST /api/v2/barong/organization/account

Add existing user into organization/subunit
POST /api/v2/barong/organization/user

Update

Update organization info
PUT /api/v2/barong/organization/update

Update organization settings
PUT /api/v2/barong/organization/settings

Delete

Delete subunit in the organization
DELETE /api/v2/barong/organization/account

Delete user in the organization
DELETE /api/v2/barong/organization/user

app/api/v2/base.rb
```diff
@@ -65,6 +65,7 @@ class Base < Grape::API
 
     mount Management::Base => '/management'
     mount Admin::Base      => '/admin'
+    mount Commercial::Base => '/commercial'
```
chumaknadya: so here we should point to /organization

app/models/organization.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class Organization < ApplicationRecord
+  has_one  :organization,   foreign_key: :parent_id
+  has_many :organizations,  foreign_key: :parent_id
+  has_many :memberships,    dependent: :destroy
+
+  validates :oid, presence: true, uniqueness: true
+  validates :name, presence: true, uniqueness: true
```
chumaknadya: validates :oid,:name,presence: true, uniqueness: true

app/models/organization.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class Organization < ApplicationRecord
+  has_one  :organization,   foreign_key: :parent_id
```
chumaknadya: in this case maybe lets call it parent_organization

app/models/membership.rb
```diff
@@ -0,0 +1,23 @@
...
+end
+
+# == Schema Information
+# Schema version: 20210514034514
+#
+# Table name: memberships
+#
+#  id              :bigint           not null, primary key
+#  user_id         :bigint
+#  organization_id :bigint           not null
```
chumaknadya, line 21: should membership have a role too?

app/models/user.rb
```diff
@@ -190,6 +191,21 @@ def drafted_profile
     self.profiles&.find_by(state: 'drafted')
   end
 
+  attr_writer :current_organization
+
+  def organization
+    return @current_organization unless @current_organization.nil?
+
+    return nil if memberships.empty?
+    return nil if memberships.first.organization_id.nil?
```
chumaknadya: return nil if memberships.empty? || memberships.first.organization_id.nil?

app/models/admin_switch_session.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+class AdminSwitchSession < ApplicationRecord
```
chumaknadya: why do you need this class?

app/api/v2/commercial/organizations.rb
```diff
@@ -0,0 +1,467 @@
...
+          desc 'Returns array of organizations',
+               failure: [],
+               success: API::V2::Commercial::Entities::Organization
+          params do
+            use :pagination_filters
+          end
+          get do
+            admin_organization_authorize! :read, Organization
+
+            organizations = Organization.where(parent_id: nil)
```
chumaknadya: lets do a scope for this

app/api/v2/commercial/organizations.rb
```diff
@@ -0,0 +1,467 @@
...
+            requires :group,
+                     type: String,
+                     allow_blank: false,
+                     desc: 'organization fee group'
+          end
+          post do
+            admin_organization_authorize! :create, Organization
+
+            declared_params = declared(params, include_missing: false)
+            org_params = declared_params.slice('name', 'group')
```
chumaknadya: why do you need slice here?

app/api/v2/commercial/organizations.rb
```diff
@@ -0,0 +1,467 @@
...
+                     type: String,
+                     allow_blank: false,
+                     desc: 'organization name'
+            requires :group,
+                     type: String,
+                     allow_blank: false,
+                     desc: 'organization fee group'
+          end
+          post do
+            admin_organization_authorize! :create, Organization
```
chumaknadya: please do `::Organization` in all places where you have authorize

app/api/v2/commercial/organizations.rb
```diff
@@ -0,0 +1,467 @@
...
+                end
+              end
+              organization.destroy
+              status 200
+            end
+          end
+        end
+      end
+    end
+  end
```
chumaknadya: could we somehow split this file by several files, cause its quite a  lot of logic in one file

app/api/v2/entities/organization.rb
```diff
@@ -0,0 +1,33 @@
...
+        expose :id,
+               documentation: {
+                 type: 'Integer',
+                 desc: 'Organization ID'
+               }
+
+        expose :oid,
+               documentation: {
+                 type: 'String',
+                 desc: 'Organization OID'
```
chumaknadya, line 22: do we need more params here?
like address,city...

chumaknadya (changes requested):

app/api/v2/commercial/accounts.rb
```diff
@@ -0,0 +1,59 @@
...
+          get do
+            # Check account in the organization that user belong to
+            members = Membership.joins('LEFT JOIN organizations ON organizations.id = memberships.organization_id')
+                                .where(user_id: current_user.id)
+                                .select('memberships.*,organizations.name, organizations.parent_id')
+                                .pluck(:organization_id, :'organizations.name', :'organizations.parent_id')
+                                .map { |id, name, pid| { id: id, name: name, pid: pid } }
+
+            # Verify barong admin has AdminSwitchSession ability
+            if admin_organization? :read, AdminSwitchSession
```
chumaknadya: if you have admin_organization, it means only admin should have access to it
Then it should be on admin filder

app/api/v2/commercial/accounts.rb
```diff
@@ -0,0 +1,59 @@
...
+                 { code: 404, message: 'Account not found' }
+               ],
+               success: API::V2::Commercial::Entities::Account
+          params do
+            use :pagination_filters
+          end
+
+          get do
+            # Check account in the organization that user belong to
+            members = Membership.joins('LEFT JOIN organizations ON organizations.id = memberships.organization_id')
```
chumaknadya: do this as scope, i mean .joins()

app/models/membership.rb
xenirio: As I talked with Jeng. Camille need to use role in user instead. So, I just removed.

app/models/admin_switch_session.rb
xenirio: This is for `admin_organization? :read, AdminSwitchSession`. Not sure is there better way to check with string?

chumaknadya (changes requested):

app/api/v2/commercial/entities/account.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    class Account < AccountBase
+      expose :uids do |member|
+        member.memberships.map { |m| m.user.uid }
+      end
+    end
+
+    class OrganizationAccount < AccountBase
+      expose :users do |member|
```
chumaknadya: why we call field  users, and return length?

chumaknadya (changes requested):

app/api/v2/commercial/accounts.rb
xenirio: Actually, Only admin who have ability `AdminSwitchSession` and any Organization member can access this endpoint.
We use the same endpoint for both role.

app/api/v2/entities/organization.rb
xenirio: I do as base class for Organization. It could be expose more of details in `app/api/v2/commercial/entities`

app/api/v2/commercial/entities/account.rb
xenirio: This model for organization management for Tower to show users in each subunit

image

app/models/membership.rb
xenirio: As I've updated. We will user membership role with `SessionSwitch` ability case. I will bring it back. Thanks!

app/api/v2/identity/sessions.rb
calj: then can you handle this case first?
it's better to avoid negative conditions to make the code more readable.

PR #1317, opened 2021-05-26
Fix: add extension for document

akushniruk:


app/api/v2/entities/document.rb
```diff
@@ -11,7 +11,8 @@ class Document < API::V2::Entities::Base
                 desc: 'File url'
```
chumaknadya: change desc here

chumaknadya (changes requested):

chumaknadya:
@akushniruk Are you sure we are not using this endpoint to get array of documents?
I just wondering cause we call .upload function here

